PR #69, opened 2022-02-18
`esbuild`

twrichards:
Migrate away from [`@vercel/ncc`](https://www.npmjs.com/package/@vercel/ncc) for bundling all the lambdas, and instead use [`esbuild`](https://esbuild.github.io/) as it's lightning fast.

### DIFF
--- a/workflow-bridge-lambda/src/index.ts
+++ b/workflow-bridge-lambda/src/index.ts
@@ -16,7 +16,14 @@ async function getPinboardByComposerId(composerId: string) {
   const contentResponse = await fetch(
     `${WORKFLOW_DATASTORE_API_URL}/content/${composerId}`
   );
-  const data = (await contentResponse.json()).data;
+  const data = ((await contentResponse.json()) as {
+    data: {
+      externalData: {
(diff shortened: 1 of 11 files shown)


workflow-bridge-lambda/src/index.ts
```diff
@@ -16,6 +16,7 @@ async function getPinboardByComposerId(composerId: string) {
   const contentResponse = await fetch(
     `${WORKFLOW_DATASTORE_API_URL}/content/${composerId}`
   );
+  // @ts-ignore
```
andrew-nowak, line 19: what's being ignored, is there any way to get them fixed? it's a bit of a shame to have these warnings coming in

andrew-nowak (approved):

workflow-bridge-lambda/src/index.ts
twrichards: factored these out (for posterity it was ignoring `TS2571: Object is of type 'unknown'` on the await beneath)

andrew-nowak (approved):

PR #64, opened 2022-02-16
Create CI action on github

andrew-nowak:
Co-authored-by: @twrichards 

## What does this change?

Move to Github actions for pinboard CI

## How to test

Do the commits on this PR build in Github?

## How can we measure success?

- Actions config is included in VCS.
- Speedier builds than Teamcity

PR #70, opened 2022-02-18
Split out `ci.sh` script steps into GitHub Actions workflow steps

twrichards:
Now that we have a fully version controlled build process (GitHub Actions as opposed to TC - see #64) it would be nice if we could have easy visibility on the different steps involved in CI directly in the workflow output, notably inc. the time breakdown...
image


scripts/ci.sh
```diff
@@ -1,38 +0,0 @@
-#!/bin/bash
-set -e
-
-# https://classic.yarnpkg.com/en/docs/cli/policies/#toc-policies-set-version
-# This could be removed if all TC agents guaranteed global yarn installed
-shopt -s expand_aliases
-PROJECT_ROOT=$(pwd)
-alias yarn='node ${PROJECT_ROOT}/.yarn/releases/yarn-*.*js'
```
andrew-nowak, line 8: If this has been removed, do we still need the vendored version of yarn in this repo?

andrew-nowak (approved):

PR #75, opened 2022-02-21
gzip client bundle

twrichards:
This turns on compression (most likely gzip) at the API Gateway layer (to avoid the need for build time compression using webpack plugins etc, and extra logic in `bootstrapping-lambda`). 

This halves the load time for the main client bundle from around one second, to around half a second 🎉  ⚡ .

### DIFF
--- a/cdk/stack.ts
+++ b/cdk/stack.ts
@@ -567,6 +567,7 @@ export class PinBoardStack extends Stack {
         deployOptions: {
           stageName: "api",
         },
+        minimumCompressionSize: 0, // gzip responses where the client (i.e. browser) supports it (via 'Accept-Encoding' header)
       }
     );
 


bootstrapping-lambda/src/server.ts
```diff
@@ -91,13 +91,17 @@ server.get("/pinboard.loader.js", async (request, response) => {
...
+      `${mainJsFilename}.gz`;
+
     response.send(
       loaderTemplate(
         {
           appSyncConfig,
           userEmail: maybeAuthedUserEmail,
         },
-        mainJsFilename,
+        maybeGzippedMainJsFilename || mainJsFilename,
```
twrichards, line 104: we probs also need to send a `Content-Encoding: gzip` header in the response

twrichards: done

andrew-nowak (approved):

PR #81, opened 2022-02-24
Refactor floaty

andrew-nowak:
## What does this change?

Split up and simplify floaty:

- move out logic into new component "GlobalState"
- move out panel styling into Panel
- passing state and methods around through a new context rather than passing large volumes of props
- pull the shadow dom root up to encompass both the floaty and the panel

## How to test

Check for regressions. Tested on CODE.

## Accessibility


client/src/globalState.tsx
```diff
@@ -0,0 +1,274 @@
...
+  payloadToBeSent: PayloadAndType | null;
+  clearPayloadToBeSent: () => void;
+  isExpanded: boolean;
+  setIsExpanded: (_: boolean) => void;
+  userLookup: { [email: string]: User } | undefined;
+  hasWebPushSubscription: boolean | null | undefined;
+  showNotification: (item: Item) => void;
+  clearDesktopNotificationsForPinboardId: (pinboardId: string) => void;
+}
+export const GlobalState: React.FC<GlobalStateProps> = ({
```
twrichards, line 70: this name is confusing at the call site in `app.tsx`, what do you think of...
```suggestion
export const FloatyAndPanelWithGlobalState: React.FC<GlobalStateProps> = ({
```

client/src/styling.ts
```diff
@@ -0,0 +1,15 @@
+import { css } from "@emotion/react";
+import { space } from "@guardian/source-foundations";
+
+export const bottom = 108;
+export const right = 15;
+export const floatySize = 40;
+export const boxShadow =
+  "rgba(0, 0, 0, 0.14) 0px 0px 4px, rgba(0, 0, 0, 0.28) 0px 4px 8px";
+export const standardFloatyContainerCss = css`
```
twrichards, line 9: this refers to the scrollable panel right?
```suggestion
export const standardPanelContainerCss = css`
```

client/src/pinboard.tsx
```diff
@@ -146,8 +149,8 @@ export const Pinboard = ({
         <HeadingPanel
```
twrichards, line 149: to avoid confusion with new `Panel` component
```suggestion
        <Header
```
or
```suggestion
        <PanelHeading
```

client/src/globalState.tsx
```diff
@@ -0,0 +1,274 @@
...
+  };
+
+  return (
+    <GlobalStateContextProvider value={contextValue}>
+      <Floaty
+        presetUnreadNotifications={presetUnreadNotifications}
+        isExpanded={isExpanded}
+        setIsExpanded={setIsExpanded}
+        hasError={hasError}
+        hasUnread={hasUnread}
```
twrichards, line 264: if we're exposing so much via the context, is there any need to pass these as props?

twrichards: if we do it all via the context, we can avoid `GlobalState` having to combine the Floaty and Panel, and instead just export the Provider, then combine them in `app.tsx` - avoiding the need for the gross name in  https://github.com/guardian/editorial-tools-pinboard/pull/81#discussion_r816869493

client/src/pinboard.tsx
andrew-nowak: will do this in #82

client/src/styling.ts
andrew-nowak: included in #82

client/src/globalState.tsx
```diff
@@ -0,0 +1,285 @@
...
+  hasUnread: boolean;
+  hasUnreadOnOtherPinboard: (pinboardId: string) => boolean;
+
+  presetUnreadNotificationCount: number | undefined;
+}
+const GlobalStateContext = React.createContext<GlobalStateContextShape | null>(
+  null
+);
+
+const GlobalStateContextProvider = GlobalStateContext.Provider;
```
twrichards, line 52: super minor, but we don't really need to alias this now that its only being used in this component

client/src/globalState.tsx
```diff
@@ -0,0 +1,285 @@
...
+    isExpanded,
+    setIsExpanded,
+
+    presetUnreadNotificationCount,
+  };
+
+  return (
+    <GlobalStateContextProvider value={contextValue}>
+      {children}
+    </GlobalStateContextProvider>
```
twrichards, line 283:
```suggestion
    <GlobalStateContext.Provider value={contextValue}>
      {children}
    </GlobalStateContext.Provider>
```

twrichards (approved):

twrichards:
I think this PR introduces a regression w.r.t. font size...
| Branch | Composer | Grid |
| --- | --- | --- |
| `main` | image  | image | 
| `an/floaty-refactor` | image | image |

... possibly related to shadow DOM

andrew-nowak:
@twrichards good spot! I'd propose to fix that when implementing the relevant design, since pinboard is not yet in production

client/src/pinboard.tsx
andrew-nowak: resolved in #82 (renamed to navigation to align with the design name)

PR #85, opened 2022-03-02
add Sentry

twrichards:
Co-Authored-By: @andrew-nowak 

## What does this change?
This PR adds Sentry to pinboard, so we can have visibility on client side errors.

Since some/all of our host applications (e.g. composer, grid) already use Sentry (but of unknown versions, and pointing at their respective Sentry projects) we needed to ensure errors originating from pinboard end up in the pinboard Sentry project. Unfortunately we can't just call `Sentry.init` since that will take over globally and we'd then recieve all errors from the host application in the pinboard Sentry project, as sentry-javascript doesn't really support multiple DSNs at once - so the implementation is a little complicated (via a Sentry global event processor which intercepts all Sentry events and we check if the stack trace originates from `pinboard.main` and if so we send explicitly to a pinboard specific Sentry client and prevent the event from going to the global sentry client i.e. the host application's Sentry project page).

Since all Apollo (GraphQL) errors are caught we also explicitly send these to the pinboard specific Sentry client.

This does increase the bundle size quite significantly - but probably worth it...
| Before | After |
| --- | --- | 
| image | image |

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,11 @@ jobs:
       id-token: write
       contents: read
     steps:
+      - name: calculate build number
+        run: |
+          LAST_TEAMCITY_BUILD=650
+          echo "BUILD_NUMBER=$(( $GITHUB_RUN_NUMBER + $LAST_TEAMCITY_BUILD ))" >> $GITHUB_ENV
+
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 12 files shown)


andrew-nowak (approved):
lgtm, working on composer - not for this PR but CSP on grid needs updating to support this - 
pinboard-grid-csp
(I'm hoping this is the cause of the [object Object] error messages and the real messages will come back when that's fixed)

PR #92, opened 2022-03-07
workflow server side search

twrichards:
## What does this change?
- Passes the existing ` searchText` (from client-side filtering) into the the `listPinboards` GraphQL Query, making use of [`apollo-link-debounce`](https://www.npmjs.com/package/apollo-link-debounce) to avoid too frequent calls (on every key-press).
- via a new `searchText` input on the `listPinboards` GraphQL Query (in the `schema.graphql`)
- this`searchText` argument is then recieved by `workflow-bridge-lambda` and passed to the `workflow` API call via the (existing) `text` query parameter (which searches across notes, headline and working title - see https://github.com/guardian/workflow/blob/b95c45e1669f37765c5defc5b11ecafcc5212507/datastore/app/query/Query.scala#L81-L85)

## How to test
With this branch deployed to CODE (since local uses CODE AppSync) - search for various pinboards and observe the network requests in the Network tab of Developer Tools.

This is a necessity for PROD, to allow us to find pinboards amongst the 9k plus results (without having to return them all to the client).

### DIFF
--- a/client/src/entry.tsx
+++ b/client/src/entry.tsx
@@ -13,8 +13,11 @@ import * as SentryHub from "@sentry/hub";
 import { onError } from "@apollo/client/link/error";
 import { GIT_COMMIT_HASH } from "../../GIT_COMMIT_HASH";
 import { BUILD_NUMBER } from "../../BUILD_NUMBER";
+import DebounceLink from "apollo-link-debounce";
 
 const SENTRY_REROUTED_FLAG = "rerouted";
+
+const DEFAULT_APOLLO_DEBOUNCE_DELAY = 0; // zero in-case used by mistake
 
 export function mount({
   userEmail,
@@ -100,6 +103,7 @@ export function mount({
 
   const apolloClient = new ApolloClient({
     link: ApolloLink.from([
+      new DebounceLink(DEFAULT_APOLLO_DEBOUNCE_DELAY), // order is important
       apolloErrorLink,
       createAuthLink(apolloUrlInfo),
       createSubscriptionHandshakeLink(apolloUrlInfo),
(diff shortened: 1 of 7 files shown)


client/src/entry.tsx
```diff
@@ -100,6 +101,7 @@ export function mount({
 
   const apolloClient = new ApolloClient({
     link: ApolloLink.from([
+      new DebounceLink(0), // order is important
```
andrew-nowak, line 104: why 0?

andrew-nowak: because this is the default debounce timeout, the queries that want a debounce will set it, eg. `gqlListPinboards` in `SelectPinboard` - may be worth adding as a comment, or putting the 0 in a named variable?

workflow-bridge-lambda/src/index.ts
```diff
@@ -37,13 +39,15 @@ async function getPinboardByComposerId(composerId: string) {
   return { ...data, status: data.externalData?.status };
 }
 
-async function getAllPinboards() {
+async function getAllPinboards(searchText?: string) {
   const fields = ["id", "title", "composerId"].join(",");
 
+  const searchQueryParamClause = searchText ? `&text=${searchText}` : "";
```
andrew-nowak, line 45: does `searchText` need URL encoding?

client/src/entry.tsx
twrichards: good point, thanks - done

workflow-bridge-lambda/src/index.ts
twrichards: although it didn't cause any problems so far, still a good idea - done

PR #96, opened 2022-03-09
Implement chat window designs

andrew-nowak:
## What does this change?

Implements designs for the chat panel, listed in https://trello.com/c/JtRZoPxe/378-pinboard-implement-chat-window-design

Some elements in the design have not been implemented as the
functionality they would demonstrate has not been implemented:

 - "'asset' was added" messages
 - Typing cues
 - Colouring URLs in messages
 - Highlighting unread mentions of 'me'
 
## Images

image

image

image

image

### DIFF
--- a/client/src/avatarRoundel.tsx
+++ b/client/src/avatarRoundel.tsx
@@ -1,9 +1,9 @@
 import { css } from "@emotion/react";
 import { neutral } from "@guardian/source-foundations";
-import { SvgPerson } from "@guardian/source-react-components";
 import React from "react";
 import { User } from "../../shared/graphql/graphql";
 import { composer } from "../colours";
+import { agateSans } from "../fontNormaliser";
 
 interface AvatarRoundelProps {
   maybeUser: User | undefined;
@@ -35,18 +35,24 @@ export const AvatarRoundel = ({
     />
   ) : (
     <span
+      title={tooltip}
       css={css`
         width: ${size}px;
         height: ${size}px;
         border-radius: 50%;
         background-color: ${composer.primary[300]};
-        fill: ${neutral[100]};
+        color: ${neutral[100]};
         display: flex;
+        user-select: none;
         justify-content: center;
         align-items: center;
+        ${size < 20 // arbitrary breakpoint
+          ? agateSans.xxsmall()
+          : agateSans.small()}
       `}
     >
-      <SvgPerson size="xsmall" />
+      {(maybeUser?.firstName || userEmail).charAt(0).toUpperCase()}
+      {maybeUser?.lastName?.charAt(0).toUpperCase()}
     </span>
   );
 };
--- a/client/src/avatarRoundel.tsx
+++ b/client/src/avatarRoundel.tsx
@@ -1,6 +1,9 @@
 import { css } from "@emotion/react";
+import { neutral } from "@guardian/source-foundations";
+import { SvgPerson } from "@guardian/source-react-components";
 import React from "react";
 import { User } from "../../shared/graphql/graphql";
+import { composer } from "../colours";
 
 interface AvatarRoundelProps {
   maybeUser: User | undefined;
@@ -31,24 +34,19 @@ export const AvatarRoundel = ({
       src={maybeUser?.avatarUrl}
     />
   ) : (
-    <div
-      key={userEmail}
-      title={tooltip}
+    <span
       css={css`
-        background-color: #586293;
-        border-radius: 50%;
-        color: white;
         width: ${size}px;
         height: ${size}px;
+        border-radius: 50%;
+        background-color: ${composer.primary[300]};
+        fill: ${neutral[100]};
         display: flex;
-        justify-content: space-around;
+        justify-content: center;
         align-items: center;
-        font-size: ${size / 2}px;
-        line-height: ${size / 2}px;
       `}
     >
-      {(maybeUser?.firstName || userEmail).charAt(0).toUpperCase()}
-      {maybeUser?.lastName?.charAt(0).toUpperCase()}
-    </div>
+      <SvgPerson size="xsmall" />
+    </span>
   );
 };
--- a/client/src/createItemInputBox.tsx
+++ b/client/src/createItemInputBox.tsx
@@ -101,6 +101,7 @@ export const CreateItemInputBox = ({
         setMessage(event.target.value);
       }}
       onKeyPress={(event) => {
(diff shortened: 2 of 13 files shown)


client/src/styling.ts
```diff
@@ -13,3 +13,35 @@ export const standardPanelContainerCss = css`
...
+    border: none;
+    cursor: default;
+  }
+  &::-webkit-scrollbar-thumb {
+    background-color: ${fg};
+    background-clip: padding-box;
+    border: 2px solid transparent;
+    border-radius: 999px;
+  }
+`;
```
andrew-nowak, line 47: chrome and firefox don't use the same css props to set scrollbars - firefox uses `scrollbar-color` and `scrollbar-width` which only offer very limited styling. setting them removes the colour change on hover/click and there's no way (as far as I can tell) to set them. Chrome's `-webkit-scrollbar`s are much more flexible, eg. they'll respect `:hover` pseudoclasses. I think the only way to get hover styles back on firefox is to remove the `scrollbar-color` and let firefox pick the stylings (we can keep `scrollbar-width` though!). That would look like this:

image

twrichards:
This is looking 🔥 

I will do a proper code review tomorrow, tiny thing I spotted from your screenshot 
image - note how the bottom of the `@` is clipped. I think we can raise this up, happy to pair...

andrew-nowak:
@twrichards That's not being clipped; that's how the `@` glyph looks in Agate Sans! It looks the same in Composer and Grid

twrichards:
> @twrichards That's not being clipped; that's how the `@` glyph looks in Agate Sans! It looks the same in Composer and Grid
> 
> image
> 
> image

omg @andrew-nowak - that's depressing!

what do you think @paperboyo and @pradasa ? shall we use a different guardian typeface just for the `@` - I'm not sure I can look at that every day when coding!

paperboyo:
It all looks great! Including the `@` even if it lacks a bit of jaw… (I would **absolutely not** load another font, would be a nightmare matching kernings italics, bolds etc)

image

client/src/avatarRoundel.tsx
```diff
@@ -31,24 +34,19 @@ export const AvatarRoundel = ({
...
+        justify-content: center;
         align-items: center;
-        font-size: ${size / 2}px;
-        line-height: ${size / 2}px;
       `}
     >
-      {(maybeUser?.firstName || userEmail).charAt(0).toUpperCase()}
-      {maybeUser?.lastName?.charAt(0).toUpperCase()}
-    </div>
+      <SvgPerson size="xsmall" />
```
twrichards, line 49: one for @pradasa (as implementation is fine) - sad we're losing the initials, not that many people have a profile pic set and people are familiar with initials from presence etc.

client/src/seenBy.tsx
```diff
@@ -1,9 +1,16 @@
 import { css } from "@emotion/react";
+import { palette } from "@guardian/source-foundations";
+import { SvgPlus } from "@guardian/source-react-components";
 import React from "react";
 import { LastItemSeenByUser, User } from "../../shared/graphql/graphql";
+import { agateSans } from "../fontNormaliser";
 import { AvatarRoundel } from "./avatarRoundel";
 import { formattedDateTime } from "./util";
 
+const maxSeenByIcons = 2;
```
twrichards, line 10: @pradasa having a max seems sensible, 2 seems a too few, IMO

client/src/seenBy.tsx
```diff
@@ -15,25 +22,52 @@ export const SeenBy = ({ seenBy, userLookup }: SeenByProps) => (
...
+        />
+      </div>
     ))}
+    {seenBy.length > maxSeenByIcons && (
+      <div
+        css={css`
+          transform: translateX(-${maxSeenByIcons * roundelOverlapPct}%);
+        `}
+      >
+        <span
```
twrichards, line 59: perhaps we could have a `title` tooltip, showing the email addresses of all the `seenBy`s that didn't make the cut

twrichards:
> It all looks great! Including the `@` even if it lacks a bit of jaw… (I would **absolutely not** load another font, would be a nightmare matching kernings italics, bolds etc)

Fair enough, the `Typography Tsar`™️ @paperboyo  has spoken, I can't argue, plus when actually using it's not as jarring as I first thought.

client/src/createItemInputBox.tsx
```diff
@@ -92,34 +94,46 @@ export const CreateItemInputBox = ({
...
+        setMessage(event.target.value);
+      }}
+      onKeyPress={(event) => {
+        if (isEnterKey(event)) {
+          if (message) {
+            sendItem();
+          }
+          event.preventDefault();
+        }
+      }}
```
twrichards, line 110: although logically this hasn't changed (nice readability improvement btw), I spotted two problems when using in the grid...

1. pressing enter with a payload attached, but no message (which is valid and works when clicking the send button) - doesn't send - but I think it should - from the code, I assume this has always been a problem
2. grids own keyboard shortcuts seem to be firing whilst typing, e.g. `f` makes it go fullscreen - I suspect some use of `event.stopPropagation()` at the top of the event handler regardless of what key was pressed should probably fix that

twrichards (commented):
This is absolutely glorious. Top work @andrew-nowak.

I've taken it for spin in CODE and it's almost all working nicely, l've left comments on relevant part.

Plus a few other non-blocking comments too.

client/src/createItemInputBox.tsx
andrew-nowak: Good spots! These are fixed

client/src/seenBy.tsx
andrew-nowak: great idea, done!

client/src/seenBy.tsx
pradasa: I would suggest having 2 and a + sign if we need to display 3 or more avatars. It might get noisy otherwise. 
https://www.figma.com/file/EZZWGDNgPX0s4RUMZTVs4k/%F0%9F%93%8Cchat?node-id=104%3A2978

client/src/avatarRoundel.tsx
andrew-nowak: Put the initials back as the fallback

image

twrichards (approved):
🔥

andrew-nowak:
TODO: there is some circumstance in the time formatter that produces `2 minu` (1 min and 3 min behave as expected) 
image

PR #102, opened 2022-03-15
Implement asset view

andrew-nowak:
## What does this change?

Adds in the asset view

I've chosen to implement by adding a new component to control the asset view, rather than reusing `ScrollableItems` (thoough I did also spike that implementation, which is on the [an/asset-view-2](https://github.com/guardian/editorial-tools-pinboard/compare/3ca2c...an/asset-view-2) branch.)

Pros:
- Code has a minimal footprint - all of the changes are limited to this new component and the `Pinboard` component. Reusing `ScrollableItems` means that several components would need to be aware of which tab has been selected and modify their behaviour/appearance based on that. eg. There is no modification to `ItemDisplay`.
- Asset view does not interfere with the unread notifications system. Using `ScrollableItems`, if you were at the bottom of the asset view when a new text only message arrived, the unread notification dot would be dismissed, despite not being able to see the message.
- It's easier to deduplicate images in the asset view.

Cons:
- Some of the code from `ScrollableItems` is duplicated (styling, merging of items from various sources)
- Some behaviour from `ScrollableItems` has not been replicated (though could be added again in the future), such as scrolling to the bottom of the panel on load.

## Images

image

### DIFF
--- a/client/src/pinboard.tsx
+++ b/client/src/pinboard.tsx
@@ -13,6 +13,7 @@ import { SendMessageArea } from "./sendMessageArea";
 import { useGlobalStateContext } from "./globalState";
 import { css } from "@emotion/react";
 import { palette } from "@guardian/source-foundations";
+import { AssetView } from "./assetView";
 
 export interface LastItemSeenByUserLookup {
   [userEmail: string]: LastItemSeenByUser;
@@ -32,6 +33,7 @@ export const Pinboard: React.FC<PinboardProps> = ({
   panelElement,
 }) => {
   const {
+    activeTab,
     userEmail,
     userLookup,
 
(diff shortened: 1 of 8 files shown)


client/src/pinboard.tsx
```diff
@@ -151,7 +153,7 @@ export const Pinboard: React.FC<PinboardProps> = ({
           flex-grow: 1;
         `}
       />
-      {initialItems.data && (
+      {activeTab === "chat" && initialItems.data && (
```
twrichards, line 156: minor and not really for this PR, but we should probably call this discussion rather than chat

twrichards (approved):
Awesome stuff @andrew-nowak 

Completely agree with your choice to diverge from `ScrollableItems` (well reasoned in the PR description).

PR #111, opened 2022-03-16
Fix overflowing usernames and emails

andrew-nowak:
## What does this change?

Moves word breaking css rule up to top of item display, so that usernames (and emails) are included. To also encompass emails, we must use `overflow-wrap: anywhere` to encourage the browser to break the word. Ideally the word would be broken at the `@`, but there doesn't seem to be that ability in CSS.

## Images

Before: 
image

After:
image


twrichards (approved):
Nice!

PR #116, opened 2022-03-21
Add link to composer 

andrew-nowak:
## What does this change?

Adds a new button to the pinboard navigation that opens the corresponding composer piece. Only displays when there is not a preselected pinboard (ie. looking at pinboards when not in composer; if you were in composer then you would already be in the piece)

## How to test

Look through pinboards on workflow; pick a pinboard and press the button. Do you end up on the correct composer piece?

## Images

image


client/src/seenBy.tsx
```diff
@@ -44,7 +44,7 @@ export const SeenBy = ({ seenBy, userLookup }: SeenByProps) => {
       >
         Seen by
       </span>
-      {seenBy.slice(0, maxSeenByIcons).map(({ userEmail, seenAt }, i) => (
+      {seenBy.slice(0, maxSeenByIcons).map(({ userEmail }, i) => (
```
twrichards, line 47: not sure this was intentional, and I guess would break parts of https://github.com/guardian/editorial-tools-pinboard/pull/109

client/src/navigation/index.tsx
```diff
@@ -50,6 +59,7 @@ export const Navigation = (props: PropsWithChildren<NavigationProps>) => {
             onClick={props.clearSelectedPinboard}
             icon={BackArrowIcon}
             unreadCount={unreadNotificationCountOnOtherPinboard}
+            title="Open another pinboard"
```
twrichards, line 62:
```suggestion
            title="View other pinboards"
```

twrichards (commented):
Looks great, will give it a test on CODE shortly, left a few comments in the meantime

twrichards (approved):

PR #118, opened 2022-03-29
trigger `users-refresher-lambda` on permission changes

twrichards:
Previously the `users-refresher-lambda` ran on a 6 hourly schedule (to keep load on google APIs low), which kept our list of users (needed for name resolution, avatars and to populate the _mentions_ suggestions) somewhat up to date. However the recent live blog pilots highlighted the fact that we still needed to run this lambda manually after a permission change to avoid waiting up to 6 hours before users could safely use pinboard.

Since the permissions are loaded from an S3 file, it makes sense to fire our lambda when that permission file changes and update users so long as there are `pinboard` permission changes (note that the permission file contains all the permissions for all tools, so its rare that an update to this file will be for pinboard.

This means we can also reduce the scheduled frequency to once a day.

### DIFF
--- a/cdk/stack.ts
+++ b/cdk/stack.ts
@@ -11,6 +11,7 @@ import {
 } from "@aws-cdk/core";
 import * as lambda from "@aws-cdk/aws-lambda";
 import * as S3 from "@aws-cdk/aws-s3";
+import * as S3notifications from "@aws-cdk/aws-s3-notifications";
 import * as iam from "@aws-cdk/aws-iam";
 import * as ssm from "@aws-cdk/aws-ssm";
 import * as apigateway from "@aws-cdk/aws-apigateway";
@@ -25,6 +26,10 @@ import { APP, userTableTTLAttribute } from "../shared/constants";
 import crypto from "crypto";
 import { DynamoEventSource } from "@aws-cdk/aws-lambda-event-sources";
 import { ENVIRONMENT_VARIABLE_KEYS } from "../shared/environmentVariables";
+import {
+  PERMISSIONS_BUCKET,
+  PERMISSIONS_FILENAME,
+} from "../shared/permissions";
 
 export class PinBoardStack extends Stack {
   constructor(scope: Construct, id: string, props?: StackProps) {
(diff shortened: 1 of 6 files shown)


notifications-lambda/src/index.ts
```diff
@@ -4,34 +4,36 @@ import {
...
     "notifications/privateVapidKey"
   );
 
+  const items: Item[] = event.Records.reduce(
+    (acc, record) =>
+      record.dynamodb?.NewImage
+        ? [
+            ...acc,
+            AWS.DynamoDB.Converter.unmarshall(record.dynamodb.NewImage) as Item,
+          ]
```
twrichards, line 32: the official type has `dynamodb` and `NewImage` as optional, so we now `reduce` to filter those out whilst also mapping to `Item`. We also do this once now, rather than for each matching user result (was wasteful before).

cdk/stack.ts
```diff
@@ -575,16 +580,26 @@ export class PinBoardStack extends Stack {
...
 
+    S3.Bucket.fromBucketName(
+      thisStack,
+      PERMISSIONS_BUCKET,
+      PERMISSIONS_BUCKET
+    ).addObjectCreatedNotification(
+      // includes the file being updated
+      new S3notifications.LambdaDestination(usersRefresherLambdaFunction),
+      { prefix: `${STAGE}/${PERMISSIONS_FILENAME}` }
+    );
```
akash1810, line 591: Worth chatting with @jacobwinch here, who discovered CDK's S3 lambda triggers will deploy an intermediate, python based, lambda here (see also https://github.com/guardian/deploy-tools-platform/pull/579).

Would also recommend setting up [snapshot tests](https://github.com/guardian/cdk/blob/main/docs/best-practices.md#snapshot-testing) on this stack which will highlight this strangeness early on (i.e. before deploying).

twrichards: OMG you're right (see https://github.com/guardian/editorial-tools-pinboard/pull/118/commits/65888d8e2cda477727601491cbaa28bcc78dfe1e) I think I'll have to go down the SQS/SNS route as I really don't like this

twrichards: UPDATE: I paired with @jacobwinch and we deployed this branch as it stands... it turns out that the python lambda is only invoked at cloudformation time (presumably the APIs it needed to call to set up the notification on an existing bucket aren't available in CFN and so it needed this more complex runtime to execute those 🙈 ) so it's not quite as bad as we first thought (i.e. python lambda was some sort of proxy for each S3 event) although the lambda lives on beyond the cloudforming, so still messy!

FYI @akash1810 @jonathonherbert @davidfurey

twrichards:
It turns out this [S3 permissions cache file](https://github.com/guardian/permissions/pull/127) is written to frequently (every 30 seconds it appears) rather than on actual permission change - as such the approach in this PR is no good, since it will invoke our lambda far too frequently... back to the drawing board.

PR #119, opened 2022-03-29
FIX bug in `notifications-lambda`, specifically in `doesUserManuallyHavePinboardOpen`

twrichards:
…which needed to unpack the `manuallyOpenedPinboardIds` String Set 

(bug introduced in https://github.com/guardian/editorial-tools-pinboard/pull/97/files#diff-0d791228db97cb803b117ca4b2690f3920a25c1fce7e067c27c957fa15eb497e)


notifications-lambda/src/index.ts
```diff
@@ -47,45 +52,57 @@ export const handler = async (event: DynamoDBStreamEvent) => {
...
-                      errorPushing
+        userResults.Items.reduce<UserWithWebPushSubscription[]>(
+          (acc, user) =>
+            user.webPushSubscription
+              ? [
+                  ...acc,
+                  {
+                    ...user,
+                    manuallyOpenedPinboardIds:
+                      user.manuallyOpenedPinboardIds?.values, // extract items from String Set
```
twrichards, line 63: this is the important bit, everything else is either indentation or type tightening etc

andrew-nowak: does this come back as a js `Set`? (Tried reading the AWS docs which implies it is, but isn't clear). if so, values is a function https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/values

```suggestion
                      user.manuallyOpenedPinboardIds?.values(), // extract items from String Set
```

twrichards: no this comes back as some AWS construct, where `values` is the array. I've tested this PR, but feel free to deploy to CODE, to be sure. OR you can run it locally with `run.ts`

andrew-nowak (approved):

PR #120, opened 2022-03-29
FIX regression in `workflow-bridge-lambda` on lookup by composer ID

twrichards:
FIX regression/oversight introduced in https://github.com/guardian/editorial-tools-pinboard/pull/117, by ignoring the result of `getPinboardByComposerId` if `isNotFound:true` (i.e. how we used to treat null)

This fix re-instates the correct 'No pinboard' behaviour 
image

### DIFF
--- a/client/src/globalState.tsx
+++ b/client/src/globalState.tsx
@@ -120,7 +120,10 @@ export const GlobalStateProvider: React.FC<GlobalStateProviderProps> = ({
     if (!preselectedComposerId) {
       return;
     }
-    if (preselectedPinboardQuery.data?.getPinboardByComposerId) {
+    if (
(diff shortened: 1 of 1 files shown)


andrew-nowak (approved):
I can see this works, so approving. Did you consider fixing the issue on the client side, making it check the `isNotFound` flag in the response rather than a null check?

twrichards:
> I can see this works, so approving. Did you consider fixing the issue on the client side, making it check the `isNotFound` flag in the response rather than a null check?

good though @andrew-nowak - that's a nicer solution, I've changed this PR accordingly, will test shortly

andrew-nowak (approved):

PR #121, opened 2022-03-29
use `@types/aws-lambda` for lambda input events

twrichards:
Small refactor PR.

Makes sense to use `@types/aws-lambda` for lambda input events, rather than coding our own (partial) types.


jonathonherbert (approved):
LGTM! Unsure how to test, or whether this is covered by automated tests, but code looks like a no-op.

PR #122, opened 2022-03-30
add CDK snapshot tests 

twrichards:
@akash1810 rightly pointed out we'd benefit from some snapshot tests so we can see what's actually changing in the [generated] CloudFormation (see https://github.com/guardian/editorial-tools-pinboard/pull/118#discussion_r837793499)

## What does this change?
Adds basic snapshot tests to our CDK stack. 

This will also be useful for regression testing when refactoring `stack.ts` (to reduce repetition and use [`@guardian/cdk`](https://github.com/guardian/cdk) where we can).


jonathonherbert (approved):
Can see the checked-in snapshot and the test output in the GHA workflow, LGTM!

PR #123, opened 2022-03-30
run the `users-refresher-lambda` every minute 

twrichards:
https://trello.com/c/RIfC9Vqc/550-run-pinboard-user-refresher-lambda-on-permission-change

Previously the `users-refresher-lambda` ran on a 6 hourly schedule (to keep load on google APIs low), which kept our list of users (needed for name resolution, avatars and to populate the mentions suggestions) somewhat up to date. However the recent live blog pilots highlighted the fact that we still needed to run this lambda manually after a permission change to avoid waiting up to 6 hours before users could safely use pinboard.

Initially we hoped to trigger the `users-refresher-lambda` whenever permissions actually change (see https://github.com/guardian/editorial-tools-pinboard/pull/118) however without adding lots of complex infrastructure (including making changes to [guardian/permissions](https://github.com/guardian/permissions)) this is not possible.

## What does this change?
Instead, here we increase the frequency of the `users-refresher-lambda` schedule to **every minute** and with a payload of `{isProcessPermissionChangesOnly: true}` which checks there are actually pinboard permission changes and looks-up only those against the Google APIs. Note that users who have the permission removed are marked as `isMentionable: false`

We also introduce another schedule, which daily does a full run to ensure users who've left are marked as `isMentionable: false`.

Mention suggestions are now only populated with users with `isMentionable: true`. This new approach replaces the previous TTL concept (see #46 and #50) and retains user rows indefinitely, preferring a flag which is more explicit (and also makes the impending move away from Dynamo to a relational DB easier). 

## How to test
With this deployed to CODE

- change some non-pinboard permissions in CODE and see `NO CHANGE TO PINBOARD PERMISSIONS, exiting early` in the logs
- add the pinboard permission for someone, and see `DETECTED PINBOARD PERMISSIONS ADDED FOR ` in the logs (followed by the usual log lines about upserting).
- remove the pinboard permission for someone, and see `DETECTED PINBOARD PERMISSIONS REMOVED FOR ` in the logs (followed by the usual log lines about upserting).
- if left overnight (or run manually), see `FULL RUN` in the logs (followed by the usual log lines about upserting).

No dev intervention required when CP add new users to pinboard, and they can start using pinboard safely within a minute.

This will cost a touch more, but since pinboard currently costs only a few cents per month - this is OK.

### DIFF
--- a/users-refresher-lambda/src/index.ts
+++ b/users-refresher-lambda/src/index.ts
@@ -1,8 +1,4 @@
-import {
-  admin as googleAdminAPI,
-  admin_directory_v1,
-  auth as googleAuth,
-} from "@googleapis/admin";
+import { admin as googleAdminAPI, auth as googleAuth } from "@googleapis/admin";
 import { people as googlePeopleAPI } from "@googleapis/people";
 import * as AWS from "aws-sdk";
 import { standardAwsConfig } from "../../shared/awsIntegration";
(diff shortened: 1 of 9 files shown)


users-refresher-lambda/src/index.ts
```diff
@@ -12,30 +8,101 @@ import {
...
     (acc, { userId, active }) => (active ? [...acc, userId] : acc),
-    [] as string[]
+    []
   );
 
   if (!emailsOfUsersWithPinboardPermission) {
     throw Error("Could not get list of users with 'pinboard' permission.");
   }
 
+  const isNightlyFullRun = now.getHours() === 0 && now.getMinutes() === 0;
```
andrew-nowak, line 66: Would it be worth adding a parameter to allow manually running a full run in case of errors/emergencies?

twrichards: yeah I actually was thinking about something like that. Perhaps two schedules...

1. every minute, sends some `minimal: true` flag in the payload
2. every 24 hours, no payload (i.e. same as a manual run)

twrichards: done in https://github.com/guardian/editorial-tools-pinboard/pull/123/commits/c4c5d6747d89da353bc9c3b746e0572f8fcd54cd

andrew-nowak (approved):

PR #127, opened 2022-04-07
Create grid-bridge-lambda

andrew-nowak:
Co-authored-by: @twrichards @Aracho1 

## What does this change?

Adds a lambda to bridge to grid APIs. Initital usage is to run searches for collections and labels.

## How to test

Hard to exercise by itself as there are no UI changes in this PR. Deploy to CODE with #131, with guardian/grid#3713 on Grid TEST, and preview some searches (some examples are currently in the Pinboard named "NOW RENAMED", alternatively searches can be added from the local playground in #131).

Pinboard is able to connect to Grid to fetch previews for searches.

## Accessibility

### DIFF
--- a/artifact.json
+++ b/artifact.json
@@ -18,6 +18,11 @@
       "compress": "zip"
     },
     {
+      "action": "pinboard-grid-bridge-lambda",
+      "path": "grid-bridge-lambda/dist",
+      "compress": "zip"
+    },
+    {
       "action": "pinboard-users-refresher-lambda",
       "path": "users-refresher-lambda/dist",
       "compress": "zip"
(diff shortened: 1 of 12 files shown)


jonathonherbert (approved):
Tested locally and via CODE lambda, passes on the search string and also gets hold of labels and collections. One minor comment in code.

This approach works, and the alternative is a lot of work and likely not worth worrying about at present, but it feels precarious that we must divine an interpretation of our query string via the Grid, and then make a series of assumptions about the reply, without any explicit coupling with the Grid; if Grid things change the assumptions made here may break.

Nothing to be done about this at present, but if we find other applications start relying on this sort of approach, there may be enough debt accumulated to sink some time into a common resource, colocated with the Grid, that can make the dependency explicit and provide a common place for this sort of work.

PR #132, opened 2022-04-27
Parse chips out of grid query

andrew-nowak:
## What does this change?

Adds chips to the parsed parts of the grid query in grid-bridge-lambda

## How to test

Deploy to CODE, and either send searches through the appsync console or the lambda console, run locally by setting a url in `grid-bridge-lambda/run.ts`

Can we always see the chips, have they been parsed correctly? Can you find/come up with a query that fails parsing?


jonathonherbert (approved):
Tested locally with a few different chip combinations, including chips with interesting chars – output looks good, couldn't break it!

Taking a look on regex101, it looks like the three alternatives in the non-capturing group there because people tend to surround their chip searches with different sorts of quotes?

It might be worth including a comment for the regexes generally to indicate the different things they're intended to capture.

PR #139, opened 2022-05-23
Add telemetry to pinboard

aracho1:
## What does this change?
Add telemetry to Pinboard to measure user activities. This PR makes use of the existing ed tools telemetry client ([github](https://github.com/guardian/editorial-tools-user-telemetry-service)) which sends the data to central ELK. We are building the Grafana [dashboard](https://metrics.gutools.co.uk/d/b-tyvbs7z/pinboard?orgId=1&refresh=5m&var-Stage=CODE&var-Interval=$__auto_interval_Interval) for visual representation of the collected data.

## How to test
1) Run this branch either locally or deploy on CODE. 
2) Open Composer/Grid CODE. You should see the Pinboard floaty on the lower right corner. 
3) When you interact with Pinboard* such as:
- opening / closing the Pinboard floaty
- sending messages
- adding grid images
- changing notification settings
...and other activities defined in [Telemetry.ts](https://github.com/guardian/pinboard/blob/4cd16d460eff8eb2a76deae53f4dfaa85d81f347/client/src/types/Telemetry.ts), you should see in the Network tab of the developer console that the activity data gets sent to `https://user-telemetry.code.dev-gutools.co.uk/event`.

*You can create a test article in Workflow CODE or use existing test articles' Pinboards to test sending messages etc.

** You may notice that multiple `'MESSAGE_SEEN'` events get sent for the same event. This is a known 'bug' that we aim to fix in the future but should not affect our Telemetry data as we're only capturing this event once per user in the Dashboard.

Example:
image


client/src/types/Telemetry.ts
```diff
@@ -0,0 +1,44 @@
...
+  assetType?: GridAssetType;
+  notification?: PinboardNotificationSetting;
+  messageType?: GridAssetType | "message-only";
+  hasMentions?: boolean;
+  tab?: Tab;
+  composerId?: string;
+  composerSection?: string;
+}
+
+type GridAssetType = "grid-search" | "grid-crop" | "grid-original";
```
andrew-nowak, line 30: nit: this is already available as `PayloadType` in PayloadAndType.ts

client/src/payloadDisplay.tsx
```diff
@@ -49,9 +54,17 @@ export const PayloadDisplay = ({
             "application/vnd.mediaservice.kahuna.image",
             "true"
           );
+          sendTelemetryEvent?.(PINBOARD_TELEMETRY_TYPE.DRAG_FROM_PINBOARD, {
+            assetType: payloadAndType?.type,
+            tab: tab as Tab,
```
andrew-nowak, line 59: this looks ugly, and there's possibly type problems from doing this - tried looking for a better option and unfortunately the best I could find was

```suggestion
            ...(tab && { tab }),
```

which is not elegant but should ensure no problems with sending an undefined through

andrew-nowak (approved):
amazing work!!! just 2 (optional) nits, then ready to ship

PR #140, opened 2022-05-26
FIX missing messages when re-opening a closed pinboard

twrichards:
We've long noticed an odd (albeit rarely seen) bug when some recently sent/received messages don't appear after re-opening a pinboard (without having refreshed the page). A hunch led me to think it might be cache related (Apollo GraphQL library caches request by default).

So this PR forces a network request for all queries (by specifying `fetchPolicy` in the `defaultOptions` of the ApolloClient as per https://www.apollographql.com/docs/react/api/core/ApolloClient/#example-defaultoptions-object).

We also upgrade `@apollo/client` to latest for good measure (but also because there's various GH issues relating to bugs in how `defaultOptions` are applied which are fixed in more recent versions).

## How to test
Easiest in CODE. Open pinbaord when at the root/dashboard of CODE composer (or TEST grid), open a pinboard, send some messages, navigate back to the pinboard select screen, close the pinboard (the X beside the pinboard in the My Pinboards) list, then search for it and re-open and all the messages you sent should be there (they would have been missing before this fix).

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -15,7 +15,7 @@
     "test": "jest 2>&1"
   },
   "dependencies": {
-    "@apollo/client": "^3.5.5",
+    "@apollo/client": "^3.6.5",
     "@emotion/react": "^11.1.4",
     "@guardian/source-foundations": "^4.1.0",
     "@guardian/source-react-components": "^4.4.0",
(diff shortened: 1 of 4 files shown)


andrew-nowak (commented):
tested on code, lgtm

andrew-nowak (approved):

PR #141, opened 2022-05-26
FIX tiny images in asset view

twrichards:
...by removing redundant wrapper around `PayloadDisplay` in `AssetView` (missed in https://github.com/guardian/pinboard/pull/137) 

### ALSO 
- ensure ChatTab is selected when there is a 'payload to be sent' added (since it cannot be sent from the Asset tab)
- ensure ChatTab is selected when switching between pinboards, so that nobody accidentally misses messages because they're stuck in the AssetTab

### DIFF
--- a/client/src/globalState.tsx
+++ b/client/src/globalState.tsx
@@ -112,6 +112,10 @@ export const GlobalStateProvider: React.FC<GlobalStateProviderProps> = ({
 }) => {
   const [activeTab, setActiveTab] = useState<Tab>(ChatTab);
 
+  useEffect(() => {
+    payloadToBeSent && setActiveTab(ChatTab);
+  }, [payloadToBeSent]);
+
   const [getPreselectedPinboard, preselectedPinboardQuery] = useLazyQuery(
     gqlGetPinboardByComposerId
   );
--- a/client/src/assetView.tsx
+++ b/client/src/assetView.tsx
@@ -60,20 +60,10 @@ export const AssetView: React.FC<AssetView> = ({
       `}
     >
       {payloadsMap.map((payloadAndType) => (
-        <div
+        <PayloadDisplay
           key={payloadAndType.payload.embeddableUrl}
-          css={css`
-            margin: ${space[1]}px;
-            border: 1px solid ${neutral[86]};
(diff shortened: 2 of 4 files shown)


client/src/globalState.tsx
```diff
@@ -328,7 +328,7 @@ export const GlobalStateProvider: React.FC<GlobalStateProviderProps> = ({
     userEmail,
     userLookup,
 
-    activeTab,
+    activeTab: payloadToBeSent ? ChatTab : activeTab,
```
andrew-nowak, line 331: I don't like this, this stops you switching tabs entirely when you have a payload ready to send without any ui feedback. I think it'd be better to set the chat as the active tab when a payload is added but after that allow the user to switch back and forth as they want.

andrew-nowak, line 331: I don't like this, this stops you switching tabs entirely when you have a payload ready to send without any ui feedback. I think it'd be better to set the chat as the active tab when a payload is added but after that allow the user to switch back and forth as they want.

twrichards: yeah good point, I've replaced with a `useEffect` which is dependent on `payloadToBeSent` which should have the behaviour you suggested

twrichards: deployed to CODE and works as you described. It does make me wonder though if activeTab should be global state or per pinboard (since its confusing if you navigate up from a pinboard which is on asset tab then open a new pinboard it opens in asset view, which could be confusing as they look similar - one might not think there are any messages)

andrew-nowak: agreed, it should definitely at least appear to be a per-pinboard state (iirc it's a little tricky to actually do that as the navigation is sibling to the pinboards, so the state needs to live in panel which is effectively global. we should be able to get the same effect with a well-placed useEffect though)

twrichards: > agreed, it should definitely at least appear to be a per-pinboard state (iirc it's a little tricky to actually do that as the navigation is sibling to the pinboards, so the state needs to live in panel which is effectively global. we should be able to get the same effect with a well-placed useEffect though)

done that too now @andrew-nowak

andrew-nowak (approved):

PR #142, opened 2022-06-13
hash + salt user ids

andrew-nowak:
## Accessibility

### DIFF
--- a/bootstrapping-lambda/src/util.ts
+++ b/bootstrapping-lambda/src/util.ts
@@ -1,4 +1,42 @@
+import * as AWS from "aws-sdk";
 import { Response } from "express";
+import { STAGE, standardAwsConfig } from "../../shared/awsIntegration";
+import { pbkdf2Sync } from "crypto";
+
+const ssm = new AWS.SSM(standardAwsConfig);
+
+// memoise fetched salt - memo is invalidated at midnight
+// a failed fetch _is also memoised_!! this avoids lots of fetches if one fetch fails
+const saltmemo = (async () => {
+  let lastFetch = new Date();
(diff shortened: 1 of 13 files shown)


bootstrapping-lambda/src/util.ts
```diff
@@ -1,4 +1,42 @@
...
+// a failed fetch _is also memoised_!! this avoids lots of fetches if one fetch fails
+const saltmemo = (async () => {
+  let lastFetch = new Date();
+  let salt: Buffer | undefined;
+
+  const get = async () => {
+    if (lastFetch.getUTCDate() === new Date().getUTCDate()) {
+      return salt;
+    }
+    const saltParameter = ssm
```
twrichards, line 18: I think this could be done with the existing param store retrieval function https://github.com/guardian/pinboard/blob/bab6d9e3a5208541a6b96200bddacb1e3f9c93ed/shared/awsIntegration.ts#L40

twrichards (commented):
Got a notification for this PR and it intrigued me. Left a suggestion, but more generally just wondering if this is worth a whole separate lambda (i.e. more boilerplate and more infrastructure) could the salt generation be done in the user-refresher-lambda when it does its nightly 'full-run' (since it's user related, so not too confusing).
Lastly a more general question, does this nightly generation of salts mean that the resulting 'private user ids' are not consistent from day to day (i.e. we can't do unique users over any greater span than 24hours)?
Great work btw!

twrichards:
closing this since, we now have relational DB (see #161) so we can pull the stats we need from there (see https://trello.com/c/UFfdDn3U/805-pinboard-stats-page-from-db) rather than having to send any identifier to the telemetry service

PR #145, opened 2022-06-22
Add script to add users to pinboards directly

andrew-nowak:
## What does this change?

Adds a script for adding given pinboard ids to a user's manually opened pinboards list. This is handy while we have the current issues around workflow items being removed causing the pinboard to become inaccessible via the search. The pinboard IDs must be located via another means (eg. the 'messages sent by piece' graph in the current pinboard dashboard).

## How to test

Run the script, passing your username and a known pinboard ID. Does the pinboard appear in your list of opened pinboards?

Speedier and less error-prone method to recover access to pinboards during testing, until persistent access in UI is developed.

### DIFF
--- a/scripts/add_to_pinboard.sh
+++ b/scripts/add_to_pinboard.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+set -eo pipefail
+
+users="$1"
+pinboardIds="$2"
+stage="$3"
+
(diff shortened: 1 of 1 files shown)


twrichards (approved):
nice work 👏 
Note, although I wouldn't redo it, another option would've been to do this in typescript as an additional file in say `users-refresher-lambda` (to reuse some existing code and keep _'everything TS'_)

PR #148, opened 2022-08-09
positionable Floaty/Panel

twrichards:
https://trello.com/c/1uT2Mdth/194-pinboard-positionable-floaty-panel

Following feedback from users (along with our own observations/suspicions) its necessary to be able to move the pinboard floaty/panel sometimes (e.g. to see info in the grid in the right sidebar).

## Requirements
- persists between reloads (per domain, so it can be different for composer vs grid for example)
- panel sits appropriately w.r.t. the floaty, depending on which quadrant of the screen the floaty is placed
- cannot be dragged off the screen (including when the screen is resized, it can't move off the edge) 

positionable_final

### DIFF
--- a/client/src/globalState.tsx
+++ b/client/src/globalState.tsx
@@ -22,6 +22,11 @@ import type { PinboardData } from "../../shared/graphql/extraTypes";
 import { isPinboardData } from "../../shared/graphql/extraTypes";
 import type { PreselectedPinboard } from "../../shared/graphql/extraTypes";
 import { ChatTab, Tab } from "./types/Tab";
+import { ControlPosition } from "react-draggable";
+import { bottom, floatySize, right } from "./styling";
+import { useDebounce } from "./util";
+
+const LOCAL_STORAGE_KEY_EXPLICIT_POSITION = "pinboard-explicit-position";
 
 interface GlobalStateContextShape {
   userEmail: string;
@@ -65,6 +70,11 @@ interface GlobalStateContextShape {
   hasUnreadOnOtherPinboard: (pinboardId: string) => boolean;
 
   presetUnreadNotificationCount: number | undefined;
+
+  explicitPositionTranslation: ControlPosition;
+  setExplicitPositionTranslation: (newPosition: ControlPosition) => void;
+  boundedPositionTranslation: ControlPosition;
+  updateBoundedPositionTranslation: (newPosition: ControlPosition) => void;
 }
 const GlobalStateContext = React.createContext<GlobalStateContextShape | null>(
   null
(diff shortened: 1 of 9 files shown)


client/src/globalState.tsx
```diff
@@ -328,6 +338,77 @@ export const GlobalStateProvider: React.FC<GlobalStateProviderProps> = ({
...
+  const [lastResized, setLastResized] = useState<number>(0);
+
+  useEffect(() => {
+    updateBoundedPositionTranslation(explicitPositionTranslation);
+  }, [lastResized]);
+
+  const resizeCompleteHandler = useDebounce(
+    () => setLastResized(Date.now()),
+    250
+  );
```
twrichards, line 399: adding this bit of state (`lastResized`) was the only way I could make it use the latest value for `explicitPositionTranslation` when calling `updateBoundedPositionTranslation` - if I called that directly in the debounce handler it would always use the initial value for `explicitPositionTranslation`, which makes sense, but there's probably a more idiomatic/nicer way to achieve this in React.

pradasa:
It's not a blocker for release at the moment, but UX-wise it would be good to consider:
- not allowing Pinboard's floaty on top of the navigation bar at the top (at least on Grid)
- if it would be possible for the panel to move with the floaty to preview what it's going to cover in the page when the floaty is being moved around. There are some considerations here ie. the panel changes size at the moment depending on where it's placed so it might be worth discussing this a bit more before implementing because it could end up being too distracting rather than helpful for the user.

client/src/navigation/index.tsx
```diff
@@ -37,8 +37,6 @@ export const Navigation = (props: PropsWithChildren<NavigationProps>) => {
     <div
       css={css`
         background-color: ${pinboard[500]};
-        border-top-left-radius: 4px;
-        border-top-right-radius: 4px;
```
andrew-nowak, line 41: did we want to make the corners square?

twrichards: no, I was being a bit lazy and didn't want to spend too much time making the radius play nicely with the corner the callout shape was sat on (because the radius on that corner looks horrible) - but that was naughty, thanks for keeping me honest. 
I've now got it working nicely and actually made it so all three corners which don't have the callout on them...
| | **Left** | **Right** |
| --- | --- | --- |
| **Top** | image | image |
| **Bottom** | image | image |

Unfortunately `Navigation` also requires the radius, which meant passing down `isTopHalf` and `isLeftHalf` as props to `Navigation`.

twrichards:
@pradasa thank you for the suggestions.

I've addressed...
> * not allowing Pinboard's floaty on top of the navigation bar at the top (at least on Grid)

... in https://github.com/guardian/pinboard/pull/148/commits/bd9e9769bd1c086ef30721515f507f089339378a

As for...
> * if it would be possible for the panel to move with the floaty to preview what it's going to cover in the page when the floaty is being moved around. There are some considerations here ie. the panel changes size at the moment depending on where it's placed so it might be worth discussing this a bit more before implementing because it could end up being too distracting rather than helpful for the user.

... excellent point and something I had considered, as its clearly more useful that way, but implementing this one is a bit more involved and will likely be more 'expensive' (in terms of re-renders) so since it's not a blocker, will leave for a future PR.

twrichards:
@pradasa after writing...
> ... excellent point and something I had considered, as its clearly more useful that way, but implementing this one is a bit more involved and will likely be more 'expensive' (in terms of re-renders) so since it's not a blocker, will leave for a future PR.

... I thought I would just see how easy and if/how janky it would be, turns out it was easy enough and seems to be smooth (on my machine at least), so its now done in https://github.com/guardian/pinboard/pull/148/commits/c2883ce51dc6375d9b4d34766af6f04099ab0418 ...
Kapture 2022-08-10 at 22 44 46

client/src/globalState.tsx
andrew-nowak: I think it's that the useEffect adding the resize event handler never removes the handler, and doesn't declare the resizeCompleteHandler as a dependency. This works for me ->

```tsx
const resizeCompleteHandler = useCallback(() => {
  updateBoundedPositionTranslation(explicitPositionTranslation);
}, [explicitPositionTranslation]);

useEffect(() => {
  const savedExplicitPositionTranslation = JSON.parse(
    window.localStorage.getItem(LOCAL_STORAGE_KEY_EXPLICIT_POSITION) ||
      JSON.stringify({ x: 0 - right, y: 0 - bottom })
  );
  setExplicitPositionTranslation(savedExplicitPositionTranslation);
  updateBoundedPositionTranslation(savedExplicitPositionTranslation);

  window.addEventListener("resize", resizeCompleteHandler);
  return window.removeEventListener("resize", resizeCompleteHandler);
}, [resizeCompleteHandler]);
```

twrichards: after some pairing we settled on a more idomatic approach, making use of `useCallback` rather than `useState` but ensuring we don't add/remove the resize listener unnecessarily frequently - see https://github.com/guardian/pinboard/pull/148/commits/19246ac6983f527b6b0aee2efef4e0d92e12ea03

twrichards:
@andrew-nowak observed some jankiness with repositioning (i.e. not smooth), especially with large chats visible (many messages/images) - this was down to huge amount re-renders of child components, we experimented with throttling the `onDrag`, but this was choppy in its own way. 
Instead we decided to unmount the scrollable area (and all the children within) during repositioning, but preserving the scroll position (via some state, a `useMemo` and a `useEffect`) so that user experience wasn't impacted - see https://github.com/guardian/pinboard/pull/148/commits/b924a78f05145cb4ff96de0c86bb89e8637b4093

andrew-nowak (approved):

PR #150, opened 2022-08-10
replace DynamoDB Streams for the `notifications-lambda` with SQS and a AppSync pipeline resolver

twrichards:
Co-Authored-By: @andrew-nowak 

https://trello.com/c/VzkeAK2f/554-migrate-pinboard-to-relational-database

In the database ADR (specifically https://github.com/guardian/pinboard/blob/ce8309a7eb088e7ae4814f9777dab0fa620387ac/ADRs/database.md?plain=1#L39) we had stated that we could replace 'DynamoDB Streams' with an `invoke_lambda` (from within the Aurora DB engine) when we move from Dynamo to RDS. However upon further investigation/experimentation, due to the fact we must use Aurora ServerlessV1 (because it's the only thing which supports the `data-api`, which AppSync relies upon) this doesn't support attaching IAM roles and so we cannot permission the RDS cluster to invoke the lambda - putting an end to that approach. As the changes to the database ADR in this PR explain, that leaves us with two choices...

- ~add a lambda and RDS proxy between AppSync and RDS (as explained in https://aws.amazon.com/blogs/mobile/appsync-graphql-sql-rds-proxy)~ - this seems like too much infrastructure complexity (at this point)
- convert the `createItem` AppSync resolver to an AppSync 'pipeline' resolver, where the first function does the DB insert as before, then the second function invokes the lambda (and we leave the lambda to look-up what it needs to from the DB, a shame but worth it) - this is the solution we went for in this PR and can be done before any re-platforming from DynamoDB to RDS (to make that task simpler later on).

## What does this change?

- convert the existing `createItem` resolver (defined in CDK) into a 'resolver function' 
- add `notifications-lambda` as an AppSync 'data source' and add a 'resolver function' to invoke it, with the output of the insert (which includes the inserted item's generated ID)
- wire the above together into a 'pipeline' resolver, which is assigned to the `createItem` Mutation
- update the `notifications-lambda` to receive the payload sent from the new 'resolver function' above (rather than the 'DynamoDB Streams' payload) and then queue it on an SQS queue, so that the pipeline resolver can return quickly (and the user/client isn't waiting on all the notifications being sent before they know their message has been inserted)...
- ... add an SQS queue to receive items which need notifications sending, which is wired up to the `notifications-lambda`...
- ... have the `notifications-lambda` also be able to be invoked with an `SQSEvent` (containing the item that it has just queued) to then do roughly what it used to (i.e. lookup which users need to receive a push notification for that item and send out all the push notifications)

## How to test
With this branch deployed to CODE (you'll need to check the pipeline resolver gets attached OK, as AWS' cloudformation of AppSync things is buggy, see https://github.com/aws/aws-appsync-community/issues/146#issuecomment-820290287 for example)...

- ensure you've subscribed to desktop notifications in pinboard (you'll see a button for it at the top of the pinboard panel if you haven't already subscribed)
- send a message mentioning yourself (mention people by typing `@` and selecting yourself)
**- ... you should still receive a desktop notification**

This is kind of a no-op (functionality wise) but removes this complexity from the process of re-platforming from DynamoDB to RDS (where otherwise we would've had to do the `lambda_invoke` thing along the way.

There will be a performance hit when sending a message (more so when the `notifications-lambda` is cold), but from our testing this is tolerable.

### DIFF
--- a/ADRs/database.md
+++ b/ADRs/database.md
@@ -1,5 +1,7 @@
 # ADR: database
 
+_Initially written 8th August 2012, but there are some updates._
+
 ## Current Scenario : DynamoDB
 
 When getting started with AWS AppSync (TODO add ADR for this), DynamoDB is an easy choice for persisting data and has served us very well so far, primarily for the following reasons...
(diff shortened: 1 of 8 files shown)


twrichards:
Closing, since this approach is much more convoluted and the increased response time of this pipeline resolver (when someone just wants to know their message is sent) is not really acceptable (bolstered by anecdotal reports of woeful latency overhead of the `data-api`).

PR #158, opened 2022-08-16
when sending message, lock input box and replace send button with spinner

twrichards:
When experimenting on https://github.com/guardian/pinboard/pull/150, I noticed that when there's latency sending the item, the UI is not locked 😬 - now the input box is disabled during sending and the send button is replaced with a spinner (from source)

sending_spinner

### DIFF
--- a/client/src/sendMessageArea.tsx
+++ b/client/src/sendMessageArea.tsx
@@ -12,6 +12,7 @@ import { composer } from "../colours";
 import SendArrow from "../icons/send.svg";
 import { buttonBackground } from "./styling";
 import { TelemetryContext, PINBOARD_TELEMETRY_TYPE } from "./types/Telemetry";
+import { SvgSpinner } from "@guardian/source-react-components";
 
 interface SendMessageAreaProps {
   payloadToBeSent: PayloadAndType | null;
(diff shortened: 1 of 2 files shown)


client/src/sendMessageArea.tsx
```diff
@@ -121,14 +125,18 @@ export const SendMessageArea = ({
           }
         `}
         onClick={() => sendItem()}
-        disabled={!message && !payloadToBeSent}
+        disabled={(!message && !payloadToBeSent) || isItemSending}
```
davidfurey, line 128: This expression is starting to get a bit complicated.  Not sure if:
`!(message || payloadToBeSent) || isItemSending`
is any improved though.

davidfurey (approved):

client/src/sendMessageArea.tsx
twrichards: fair, I've improved it slightly in https://github.com/guardian/pinboard/pull/158/commits/938eb3d3cb20182fcf2afe8194da8f5bd458d8a6

davidfurey: Still not completely sure it is better. But it isn't worse 🤷

PR #161, opened 2022-08-22
migrate from `DynamoDB` to `RDS Postgres`

twrichards:
https://trello.com/c/VzkeAK2f/554-migrate-pinboard-to-relational-database

Co-authored-by: @aracho1  

- Updated the database ADR to describe our experiments with Aurora Serverless V1 & V2 and instance-based, settling on a traditional RDS Postgres instance.
- Added RDS Postgres instance with an [RDS Proxy](https://aws.amazon.com/rds/proxy/) (using IAM Auth) - with security group to allow access to the RDS proxy for a number of the things listed below.
- Added new lambda `database-bridge-lambda` (similar concept to `workflow-bridge-lambda` and `grid-bridge-lambda`) to interact with the RDS proxy and run SQL Queries/Updates/Inserts. This is wired into AppSync (replacing the convoluted/proprietary dynamo resolvers).
- Added some abstractions and types for the list of queries/mutations, to make the creation of the above resolvers automatic and so the linter/TS can ensure  (at build time) we have a handler in `database-bridge-lambda` for each.
- Simplified our GraphQL Schema to flatten the input/output for some of the queries and mutations (now that we're not using Dynamo)
- Added a database 'jump host' mechanism (Autoscaling group with single instance (which self-terminates after a period of time  - in future will detect idle), some logic in `shared` to scale up the ASG and create a tunnel to the DB via the EC2 box, an alarm for when the EC2 box has been running for 12 hours i.e. self-termination has failed - so we don't waste money)
- Added a series of sql files for setting up the DB (tables, indexes, trigger etc), orchestrated by a `runDatabaseSetup.ts` script (run with `yarn database-setup` at the top-level of the project)
- Added a `runDatabaseMigration.ts` (run with `yarn database-migrate` at the top-level of the project) to migrate data from Dynamo to RDS, whilst importantly converting timestamps in epoch seconds to ISO format and mapping old-style GUID Item IDs (because that's all AppSync could create for Dynamo) to sequential numeric IDs (which postgres manages).
- Refactored `users-refresher-lambda` to read/write from the RDS DB (directly) instead of Dynamo (the rest of the logic for interacting with Google Directory APIs etc. remained untouched)
- Added a DB trigger (AFTER INSERT into the Item table) to query the User table to fund the eligible users to notify and then invoke the `notifications-lambda` (see [Invoking an AWS Lambda function from an RDS for PostgreSQL DB instance](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/PostgreSQL-Lambda.html)) PLUS obviously refactoring `notifications-lambda` to receive the item and list of users and simply to fire out the push notifications (unlike before where it had to query the Dynamo User table instead).
- Added `firstName` and `lastName` to the push notifications (rather than displaying the sender's email up to the `@`)
- Removed the Dyanmo tables (they are in fact 'detached' and not deleted - owing to the default DeletionPolicy of 'Retain') - they will be deleted manually in a few days/weeks.


client/src/push-notifications/serviceWorker.ts
```diff
@@ -3,16 +3,20 @@ import { ItemWithParsedPayload } from "../types/ItemWithParsedPayload";
...
     clientId?: string;
+    firstName?: string;
+    lastName?: string;
   }
 ) => {
+  const user =
+    item.firstName && item.lastName
+      ? `${item.firstName} ${item.lastName}`
+      : item.userEmail.substring(0, item.userEmail.indexOf("@"));
+
```
phillipbarron, line 14: If this is just for display we could attempt to format the name section of the email with something like 
`item.userEmail..split("@")[0].split(".").map(name  => name.charAt(0).toUpperCase() + name.slice(1)).join(" ")`

ADRs/database.md
```diff
@@ -34,6 +34,21 @@ Assuming we stick with AWS AppSync (managed GraphQL) - which we are as it provid
 
 With RDS Aurora being the only option, this leaves us with the choice of MySQL vs Postgres from the DB engine - so we're choosing Postgres mainly based on familiarity in the team (composer & workflow DBs are Postgres).
 
-## Chosen Solution : RDS Aurora (Postgres flavour)
+## Chosen Solution : RDS Postgres
```
aracho1, line 37: Excellent documentation 👍 interesting and insightful to read

database-bridge-lambda/src/index.ts
```diff
@@ -0,0 +1,64 @@
...
+  throw Error(
+    `No handler for '${databaseOperation}' operation. @typescript-eslint/switch-exhaustiveness-check should have caught this`
+  );
+};
+
+export const handler = async (
+  payload: AppSyncResolverEvent<unknown, unknown>
+) => {
+  const sql = await getDatabaseConnection();
+  const args = payload.arguments as never;
```
aracho1, line 54: I forgot why we're using the `never` type here?

client/src/itemDisplay.tsx
```diff
@@ -137,7 +137,7 @@ export const ItemDisplay = ({
       ? item.message
       : formatMentionHandlesInText(userEmail, mentions, item.message);
 
-  const dateInMillisecs = new Date(item.timestamp * 1000).valueOf(); // the AWS timestamp is in seconds
+  const dateInMillisecs = new Date(item.timestamp).valueOf();
```
phillipbarron, line 140: just to clatify, the `item` timestamp is the item column from RDS and it's no longer in seconds?

phillipbarron (approved):

database-bridge-lambda/src/index.ts
twrichards: because we don't want to enforce a type at this point (as the args will be different depending on which graphql query/mutation it is) and `never` is the advised way to do that (rather than `any` or `object`).

PR #162, opened 2022-09-02
re-focus the text area after sending a message

twrichards:
Fairly self-explanatory, but previously after sending a message the input would lose focus (annoying)... it now re-focuses the input after sending is complete (faster to send multiple messages in succession)...

refocus


twrichards:
this has closed itself when I was rebasing to change the commit order and won't let me re-open to change the base branch to `rds` - not a huge deal

PR #163, opened 2022-09-12
ensure clicking a desktop notification opens in correct environment end expand the correct pinboard

twrichards:
Previously notification clicks would always open PROD grid/composer AND the opening by query param was not implemented (now fixed/simplified in conjunction with https://github.com/guardian/workflow-frontend/pull/375).


phillipbarron (approved):
Looks good

PR #171, opened 2022-09-14
fix-forward for 'more consistent scroll to bottom #154'

twrichards:
_replaces https://github.com/guardian/pinboard/pull/164 (that was closed by GitHub in a rebasing snafu)_ 
<hr/>

#154 wasn't as consistent as we thought, this PR fixes/simplifies by replacing the `isScrolledToBottom` function, with a piece of state to represent the same, which is more explicitly set and triggers re-render accordingly.


phillipbarron (approved):
LGTM

PR #172, opened 2022-09-15
TEMPORARY return a simple maintenance message for `/pinboard.loader.js`

twrichards:
...to allow us roll out the DB migration (Dynamo -> RDS) without breaking any of the host applications (while we're setting up the DB - see https://github.com/guardian/pinboard/pull/161


phillipbarron (approved):
👍

PR #178, opened 2022-09-21
fix first scroll

twrichards:
Despite https://github.com/guardian/pinboard/pull/154, https://github.com/guardian/pinboard/pull/171 and #169 the scroll to the bottom wouldn't always happen when a pinboard was first opened (or re-opened). This PR does away with the scroll animation when 'opening' a pinboard, and via small 100ms (not noticeable) timeout it jumps straight to the bottom - which seems to work nicely. 

Involved a little refactoring of the scrollable area ref (replacing it with a `useCallback` and `useState`, as ref's `.current` property cannot be used in the dependencies of a `useEffect` or `useCallback` - see https://medium.com/@teh_builder/ref-objects-inside-useeffect-hooks-eb7c15198780).

### DIFF
--- a/client/src/scrollableItems.tsx
+++ b/client/src/scrollableItems.tsx
@@ -1,10 +1,10 @@
 import { Item, LastItemSeenByUser, User } from "../../shared/graphql/graphql";
 import React, {
+  useCallback,
   useContext,
   useEffect,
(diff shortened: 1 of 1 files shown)


phillipbarron (approved):
LGTM

PR #179, opened 2022-09-21
re-use open composer tabs

twrichards:
If pinboard is being used in a composer piece we pre-select the pinboard (via the `<pinboard-preselect` HTML tag) and to avoid confusion we make clicking any other pinboards open in a new tab. However this is problematic if that same composer piece is already open in another tab as the user may then get into composer 'save failed' scenarios (as composer guards against writes to the same piece from different tabs).

In this PR we alert on existing matching composer tab(s) rather than opening new tabs, using `BroadcastChannel` to communicate between tabs, falling back to opening in a brand new new tab after 500ms if the desired composer tab is not found. Understandably browsers prevent tabs from bringing themselves to the fore, so that's why we need to alert rather than calling `.focus()` on the relevant window (as that does nothing).

reuse_composer_tabs

### DIFF
--- a/client/src/globalState.tsx
+++ b/client/src/globalState.tsx
@@ -220,6 +220,31 @@ export const GlobalStateProvider: React.FC<GlobalStateProviderProps> = ({
       },
     });
 
+  const [interTabChannel] = useState<BroadcastChannel>(
+    new BroadcastChannel("pinboard-inter-tab-communication")
+  );
+
+  useEffect(() => {
+    interTabChannel.onmessage = (event) => {
+      if (
+        event.data.composerId &&
+        window.parent.location.href?.includes(
+          `content/${event.data.composerId}`
+        )
+      ) {
+        // unfortunately we cannot bring this tab to the fore as browsers prevent it, so we alert to make it easier to find
+        alert(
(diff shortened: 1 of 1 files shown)


pradasa:
This looks great from a UX point of view, thanks for sorting it out.
Regarding the copy, would it be worth simplifying it?

--alert message on current tab
The composer file you want to see is already open in another tab.
 
You can see an alert message on that tab too to make it easier to find but, unfortunately, you’ll need to select the tab manually.

--alert message on requested tab
This is the composer file you wanted to open from pinboard in tab “untitled - composer article”
* (just changes composer tab to composer file)

phillipbarron (approved):
LGTM

twrichards:
> This looks great from a UX point of view, thanks for sorting it out. Regarding the copy, would it be worth simplifying it?
> 
> --alert message on current tab The composer file you want to see is already open in another tab.
> 
> You can see an alert message on that tab too to make it easier to find but, unfortunately, you’ll need to select the tab manually.
> 
> --alert message on requested tab This is the composer file you wanted to open from pinboard in tab “untitled - composer article”
> 
> * (just changes composer tab to composer file)

@pradasa done! (and GIF updated) merging now...

PR #181, opened 2022-09-26
PROD Load Testing Phase 1 of 4

twrichards:

Return basic `console.log` from `/pinboard.loader.js` in PROD only, to ensure `bootstrapping-lambda` can be called for every page load of composer/grid.


phillipbarron (approved):
lgtm 👍

PR #183, opened 2022-09-28
use only 'Google People API' for `users-refresher-lambda`

twrichards:
**_Creating this PR for visibility, but will close it immediately as it doesn't work..._**

When experimenting with https://github.com/guardian/permissions/pull/158 (which turned pinboard permission ON for everyone in `CODE`) `users-refresher-lambda` quickly attempted to ingest 1000s of new users, hitting Google's APIs hard and causing some rate limiting, which had a knock-on impact on panda login in PROD briefly (because we currently use the same service account, will be addressed in a separate PR soon).

When attempting to limit the number of users looked-up/processed in a single run of `users-refresher-lambda` (as a tactical measure to guard against the above), I got a bit carried away and ended up doing a big refactor.

Only at the last minute discovering that the People API doesn't return 'names' for anyone without a Google+ account (see https://stackoverflow.com/questions/50766097/method-people-get-does-not-return-names). I had not noticed this as I had been testing with my own user (in the API explorer, which uses your own login and does return private information for yourself 🤦 ).

There are lots of nice improvements in this PR though, which I'll use as inspiration when working on group mentions very soon (where we'll very likely use `users-refresher-lambda` to process 'group membership').

### DIFF
--- a/cdk/stack.ts
+++ b/cdk/stack.ts
@@ -555,7 +555,7 @@ export class PinBoardStack extends Stack {
       usersRefresherLambdaBasename,
       {
         runtime: LAMBDA_NODE_VERSION,
-        memorySize: 128,
+        memorySize: 256,
         timeout: Duration.minutes(15),
         handler: "index.handler",
         environment: {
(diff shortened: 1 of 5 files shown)


twrichards:
As hinted in the description, this PR is for visibility only, so closing as it doesn't work.

PR #184, opened 2022-09-28
TACTICAL limit `users-refresher-lambda` to only lookup/process 50 users in one go (to avoid Google rate limiting)

twrichards:
As mentioned in https://github.com/guardian/pinboard/pull/183; when experimenting with https://github.com/guardian/permissions/pull/158 (which turned pinboard permission ON for everyone in CODE) `users-refresher-lambda` quickly attempted to ingest 1000s of new users, hitting Google's APIs hard and causing some rate limiting, which had a knock-on impact on panda login in PROD briefly (because we currently use the same service account, [will be addressed in a separate PR soon](https://trello.com/c/ICvAsa1w/855-dont-re-use-panda-google-account-for-pinboard-users-refresher-lambda)).

This TACTICAL PR limits that to 50 in a single run (including the nightly 'full run' which kinda renders it a bit useless when we have more than 50 users), but we will soon be making significant changes to `users-refresher-lambda` when we implement group mentions (https://trello.com/c/IfRJvfZV/202-pinboard-mentions-phase-3-group-mentions).

PLUS:
- simplifies what email address we use, it now just uses the original email (from the permissions list) we use to lookup the details, rather than trying to derive the email back from the response
- ensures users with permission who are missing from Google directory still get added to DB (by inferring the `firstName` and `lastName` from the email address)
- ensures DB connection always closed regardless of success/failure outcome

### DIFF
--- a/users-refresher-lambda/src/index.ts
+++ b/users-refresher-lambda/src/index.ts
@@ -138,19 +138,27 @@ export const handler = async ({
     auth,
   });
 
+  // noinspection TypeScriptValidateJSTypes
   const usersWithPinboardPermission: Array<
     BasicUser | UserFromGoogle
   > = await Promise.all(
-    emailsToLookup.map(async (emailFromPermission) => {
+    emailsToLookup.map(async (emailFromPermission: string) => {
       const userResult = await directoryService.users
         .get({
           userKey: emailFromPermission,
         })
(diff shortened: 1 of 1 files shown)


phillipbarron (approved):
LGTM

twrichards:
✅  TESTED on `CODE`

PR #185, opened 2022-09-29
load user info async in client, rather than loading all users up-front

twrichards:
Co-authored-by: @aracho1 

https://trello.com/c/KIlzwMGx/849-load-pinboard-user-info-asynchronously-on-client

As part of https://trello.com/c/gsH0VyYU/837-pinboard-prod-load-testing we gave pinboard permission to everyone in `CODE` which once loaded into the DB (by `users-refresher-lambda` - rate limiting issues aside), we began experiencing slowness in composer (and probs. grid) as the client loads all users to build a lookup and populate the mentions autocomplete and there were now 1000s so things got slow.

This synchronous/up-front approach was arguably always temporary, but recent events mean it needs to be addressed now, and ahead of https://trello.com/c/IfRJvfZV/202-pinboard-mentions-phase-3-group-mentions.

### What's changed?

- add `searchMentionableUsers` and `getUsers` GraphQL queries with handlers in `database-bridge-lambda`
- use `searchMentionableUsers` to populate mentions autocomplete (rather than useLookup) on keypress (with debounce of 250ms to avoid unnecessary DB queries without making the UI too sluggish) - ordered by first name, but preferring users with `webPushSubscription` and/or some `manuallyOpenedPinboardIds` (i.e. people who've interacted with pinboard before)
- improved the 'loading' component for the mentions autocomplete (now we see it more as we wait for response from server)
image
- handle users with same first and last names in mentions autocomplete (by using email address as unique key)
- mentions autocomplete now also supports matching full names (i.e. with space) - which is useful now we have 1000s of users (with lots of Toms for example)
- When building the user lookup, replaced upfront call to `listUsers` with piecemeal calls to `getUsers` as and when we encounter user emails where we do not have `User` information in local state.

### DIFF
--- a/cdk/test/__snapshots__/stack.test.ts.snap
+++ b/cdk/test/__snapshots__/stack.test.ts.snap
@@ -1068,7 +1068,6 @@ type Query {
   listItems(pinboardId: String!): [Item]
   listLastItemSeenByUsers(pinboardId: String!): [LastItemSeenByUser]
   getMyUser: MyUser
-  listUsers: [User]
   searchMentionableUsers(prefix: String!): [User]
   getUsers(emails: [String!]!): [User]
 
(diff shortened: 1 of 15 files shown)


phillipbarron (approved):
LGTM

PR #187, opened 2022-10-04
load google service account details from AWS Parameter Store

twrichards:
https://trello.com/c/ICvAsa1w/855-dont-re-use-panda-google-account-for-pinboard-users-refresher-lambda

NOTE the service accounts are now for our new dedicated pinboard GCP project (rather than panda ones) which are in JSON format so could remove the nasty `p12ToPem.ts`

ALSO upped the `MAX_USERS_TO_LOOKUP_IN_ONE_RUN` limit in `users-refresher-lambda` to 1000 (from 50) now that we have a completely separate GCP account and our own rate limiting threshold of 2400 req per min (for both CODE & PROD, hence 1000 each leaving a bit of headroom)

Lastly it now handles 400 errors for `Type not supported` by simply marking them as not mentionable.


phillipbarron (approved):
LGTM 👍

PR #188, opened 2022-10-04
PROD Load Testing Phase 3 of 4

twrichards:
_Following #181 and #186 _

`/pinboard.loader.js` does everything as normal, but `loaderTemplate` doesn't call `Pinboard.mount` in PROD (instead `console.log`s accordingly)


phillipbarron (approved):
👍

PR #189, opened 2022-10-04
ensure `isMentionable` is set to `true` for those users who have their permission restored (after having had it removed)

twrichards:
ronseal


twrichards:
closing this as it will be addressed with impending overhaul of `users-refresher-lambda` for group mentions where we'll begin tracking google IDs

PR #191, opened 2022-10-05
PROD Load Testing Phase 4 of 4

twrichards:
_Following #181, #186 and #188 _

`/pinboard.loader.js` does everything as normal, and Pinboard is mounted but hidden (via css `display: none`) in PROD (to ensure APIs etc. are hit with base level of load, but nobody actually sees anything pinboardy)


aracho1 (approved):
LGTM 👍

PR #193, opened 2022-10-07
refactor `user-refresher-lambda` and begin tracking `googleID`s

twrichards:
> **Warning**
> Before merging, someone must run `yarn database-setup` (at the root of the project) and choose `PROD` and then the `add googleID column to User table` option.

`users-refresher-lambda` had become a bit unwieldy with everything in a single file, this PR splits out the Google lookup behaviour into its own files and then refactors the logic for reconciling with DB to be a bit more readable/sensible (it also provides a more obvious home for the Google syncing portion of the upcoming 'group mentions' work).

Core to this refactoring was a shift away from individual calls to the [users.get](https://developers.google.com/admin-sdk/directory/reference/rest/v1/users/get) function of the Google Admin SDK Directory API (which needed to be limited to avoid exhausting the rate limit) to now use the [users.list](https://developers.google.com/admin-sdk/directory/reference/rest/v1/users/list) function and page through (500 at a time) all the users with an `@guardian.co.uk` address and reconcile those with the users we're interested in (i.e. those with pinboard permission), which results in only a handful of API calls 😎 

As part of that refactoring we now properly detect when a user has their permission re-added (as attempted/failed to do in #189).

Lastly, we needed a way to know if a user is present in the Google directory (i.e. whether they've left the organisation) beyond them just being `isMentionable`:`false` - as this can be the case because they've had their permission removed for example - so we now store the users Google ID in a new `googleID` column of the `User` table. We also anticipate using `googleID`s in some way for the 'group mentions' implementation.

When working on this we observed the Google People API returns a non-stable photo URL for a handful of the users, which meant these users would be re-written to the DB every-time... gross. We've had a problem with very long avatar URLs elsewhere in the estate in the past, and I think these users are long serving at the org and probably set the avatars a long time ago (re-uploading their avatars will probably resolve the problem, as we noticed when working on https://github.com/guardian/frontend/pull/23162). As a workaround we download the avatar content when we detect such a difference and compare that before writing to DB (logging accordingly).

## How to test
Locally (ideally whilst this branch is NOT deployed to CODE, else it might get there first for some of the below because it runs every minute)...

...run `yarn watch` inside the `users-refresher-lambda` directory, you will be given the choice of `FULL RUN` or `process permission changes only` and then a choice of stage, choose `CODE`. 

- for `process permission changes only`
   - try first removing your pinboard permission in CODE, wait a few mins and run with this option, you should see some console output like `Permission removed for ${user.email}, so marking as not mentionable`
   - try restoring your pinboard permission in CODE, wait a few mins and run again with this option, you should see some console output like `Updating details for user ${email}`
   - try running for a final time without making any permission changes (assuming nobody else has changed pinboard permissions in that moment), you should see some console output like `NO PINBOARD PERMISSIONS CHANGED, exiting early`
   
- for `FULL RUN`, you should see a number of `...processing paged response, containing ${page?.data?.users?.length} users` and then owing to the issues explained in the Warning above you should see a see handful of users in the console in log lines like `Updating details for user ${email}`. If you really wanted to be thorough you could change your avatar in Google and run with this option and see that you get updated accordingly.

### DIFF
--- a/shared/database/local/runDatabaseSetup.ts
+++ b/shared/database/local/runDatabaseSetup.ts
@@ -42,6 +42,8 @@ const runSetupSqlFile = (sql: Sql, fileName: string) =>
           .replace("$triggerName", NOTIFICATIONS_DATABASE_TRIGGER_NAME)
           .replace("$triggerName", NOTIFICATIONS_DATABASE_TRIGGER_NAME)
       ),
+    "add googleID column to User table": () =>
+      runSetupSqlFile(sql, "008-AddGoogleIDToUserTable.sql"),
   };
 
   const allSteps = async () => {
--- a/shared/database/local/setup/008-AddGoogleIdToUserTable.sql
+++ b/shared/database/local/setup/008-AddGoogleIdToUserTable.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "User" ADD COLUMN "googleID" varchar(512);
--- a/users-refresher-lambda/package.json
+++ b/users-refresher-lambda/package.json
@@ -11,6 +11,7 @@
   },
   "devDependencies": {
     "@types/iniparser": "^0.0.29",
+    "@types/node-fetch": "2",
     "@types/node-forge": "^0.9.7",
     "aws-sdk": "^2.840.0",
     "ts-node-dev": "^1.0.0"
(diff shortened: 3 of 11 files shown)


phillipbarron (approved):
LGTM

twrichards:
Column added to `PROD` as per the Warning at top of PR description.

PR #194, opened 2022-10-18
add overall type check to `pre-commit` hook

twrichards:
It was annoying having type errors make it into CI so we now type-check in the `pre-commit` hook (like we lint etc.)


phillipbarron (approved):
LGTM 👍

PR #195, opened 2022-10-19
use `bucketSsmLookup` in riff-raff.yaml 

twrichards:
...as per https://github.com/guardian/riff-raff/pull/884

### DIFF
--- a/riff-raff.yaml
+++ b/riff-raff.yaml
@@ -32,7 +32,6 @@ deployments:
     type: aws-lambda
     parameters:
       prefixStack: false
-      bucket: workflow-dist
       fileName: pinboard-bootstrapping-lambda-api.zip
       functionNames: [pinboard-bootstrapping-lambda-]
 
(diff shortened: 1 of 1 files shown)


riff-raff.yaml
```diff
@@ -32,7 +32,7 @@ deployments:
     type: aws-lambda
     parameters:
       prefixStack: false
-      bucket: workflow-dist
+      bucketSsmLookup: true
```
akash1810, line 35: Since https://github.com/guardian/riff-raff/pull/885, `bucketSsmLookup: true` is the default, so you could omit this entirely.

akash1810 (approved):

riff-raff.yaml
twrichards: ahh yes, thanks - updated in https://github.com/guardian/pinboard/pull/195/commits/226235dc71a877e6dcc06b5054bc49ed5ca9bc89

PR #196, opened 2022-10-25
expose GraphQL operation in the query param of the request, to aid debugging

twrichards:
When debugging AppSync interactions in the client, it was painful to see which AppSync/GraphQL call was which in the Network panel of Developer Tools in the browser, because the operation was only visible after clicking into the request payload. With lots of requests its useful to be able to see which is which via the URL - this PR achieves that by intercepting the outgoing request and adding the operation name as a query param (with `_` prefix to avoid any unexpected behaviour)...

| Before | After |
| --- | --- |
| image | image |


phillipbarron (approved):
super useful 👍

PR #198, opened 2022-10-26
FIX display name on desktop notifications

twrichards:
Co-authored-by: @aracho1 

Owing to an oversight in https://github.com/guardian/pinboard/commit/c52cb60deb9a0a301943ae81c8710ae547dfe4d2 desktop notifications were always displaying with the display name (firstName, space, lastName) of the recipient rather than the author of the message.

Rather than further complicate the "AFTER INSERT" trigger on `Item` table (already quite complex) we decided to just move the 'name from email' extraction logic out of `users-refresher-lambda` and re-use in `serviceWorker.ts` for the title of the desktop notification.

We also improved this functionality along the way by adding capitalisation of first letter (including hyphenated names) - and added some tests 🙀 .


phillipbarron (approved):
👍

PR #180, opened 2022-09-23
README overhaul (PLUS overall architecture diagram)

twrichards:
This PR splits up the top-level README which was too long and moves the relevant sections into README's per-module (with updates/corrections/additions along the way) with some higher level information and links in the top-level README including this architecture diagram which explains how all the different parts work together...
architecture

PR #201, opened 2022-11-07
add README for `auth-lambda`

twrichards:
noticed that I missed `auth-lambda` in #180 - this PR puts that right with a brief README and top-level link


phillipbarron (approved):
👍

PR #202, opened 2022-11-07
run `yarn graphql-refresh` in precommit hook

twrichards:
...to ensure precommit type check doesn't fail based on the generated GraphQL TS types


phillipbarron (approved):
nice

PR #203, opened 2022-11-07
Ingest groups

phillipbarron:
Co-authored-by: @twrichards 
Co-authored-by: @aracho1 

https://trello.com/c/IfRJvfZV/202-pinboard-mentions-phase-3-group-mentions

This PR adds to the `users-refresher-lambda` (recently refactored in https://github.com/guardian/pinboard/pull/193) the ability to ingest the groups defined in the AWS Param Store (JSON object, where the key is the `shorthand` [a custom short name for the group, e.g. `audienceUK`] and the value is any email address for with that group [could be primary or alias]) - during the 'full run'.

For each of the groups in the config (using the email address to lookup) ...

- hits the https://developers.google.com/admin-sdk/directory/reference/rest/v1/groups/get to fetch the google ID, name, primary email, email aliases etc. (this required adding a new auth scope `https://www.googleapis.com/auth/admin.directory.group.readonly`)
- hits the https://developers.google.com/admin-sdk/directory/reference/rest/v1/members/list to get the members of the group

... it takes the output of these two and writes the relevant values to the new `Group` and `GroupMember` tables (see below), importantly it does this in one 'transaction' where it replaces the tables in one go (by creating new tables with `_NEW` suffix, populates them, drops the originals and renames these `_NEW` tables to the main names (to ensure the referential integrity between the two tables).

### Database changes

- add `Group` table
  - `shorthand` (a custom
  - `googleID`
  - `name` (the display name we see in gmail for example)
  - `primaryEmail` (currently just used for searching)
  - `otherEmails` (currently just used for searching)
- add `GroupMember` table
   - `groupShorthand` (with foreign key constraint to the `shorthand`  from `Group` table)
   - `userGoogleID` (can't be  foreign key constraint to the `googleID` column of `User` table as that's nullable, given some users may have left and have no Google ID)
- add `groupMentions` column to `Item` table  (array of the `shorthand`s  from `Group` table)

PR #204, opened 2022-11-07
Group mentions

phillipbarron:
Co-authored-by: @twrichards 
Co-authored-by: @aracho1 

Now that we have groups (and members) thanks to https://github.com/guardian/pinboard/pull/203 we can now surface these in the client (in the `@` autocomplete when sending a message) to allows users to mention groups, via the following changes...

- changing the return type of the `searchMentionableUsers` query from array of `User` to a new type `UsersAndGroups` (with `users` and `groups` array properties) and updated the DB query to populate the groups section when there's a match on `shorthand`, `name`, `primaryEmail` or any of the `otherEmail` - see #203 for more on these fields)
- refactoring the `mentions` `Item` type from a string array to be an array of `MentionHandle` - which is a new type containing a `label` and `isMe` property, which we calculate server-side which in turn keeps the client cleaner (NOTE we continue to store just a string array for the `mentions` column in the DB, its just transformed on the way out now).
- adding a `groupMentions` property to the `Item` type, which is also an array of `MentionHandle` as above
- update the `AvatarRoundel` to also handle Groups and render them as a circle with the member count within
- **send a 'web push'/desktop notification whenever user is a member of a group mention**
- refactor the autocomplete component (`Suggestion`) to be able to display a user or a group, where the group displays the shorthand as the main heading and the 'name' (as it appears in gmail etc.) as the subheading, with the tooltip being a list of member's email addresses. This also includes logic to detect whether the suggestion is the first of that type (user or group) and display an additional heading row...
image

### DIFF
--- a/cdk/test/__snapshots__/stack.test.ts.snap
+++ b/cdk/test/__snapshots__/stack.test.ts.snap
@@ -1068,7 +1068,7 @@ type Query {
   listItems(pinboardId: String!): [Item]
   listLastItemSeenByUsers(pinboardId: String!): [LastItemSeenByUser]
   getMyUser: MyUser
-  searchMentionableUsers(prefix: String!): [User]
+  searchMentionableUsers(prefix: String!): UsersAndGroups
   getUsers(emails: [String!]!): [User]
 
   listPinboards(searchText: String): [WorkflowStub]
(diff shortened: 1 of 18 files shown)


twrichards (approved):

twrichards:
Run relevant step of `yarn database-setup` in PROD...
image
... so that users get desktop notifications for group mentions.

PR #209, opened 2022-11-08
prevent scroll on `NotTrackedInWorkflow`

twrichards:
@aracho1 made a good spot here...

When in an article which is not tracked in workflow we display a message accordingly (encapsulated by the `NotTrackedInWorkflow` component), this re-uses some styles which included a `overflow-y: auto` which meant following #206 the other things in the `SelectPinboard` were squashing this `NotTrackedInWorkflow` component, this PR addresses that...

| Before | After | 
| --- | --- |
| image | image |

### DIFF
--- a/client/src/selectPinboard.tsx
+++ b/client/src/selectPinboard.tsx
@@ -343,6 +343,7 @@ export const SelectPinboard = ({
       <div
         css={{
           ...standardPanelContainerCss,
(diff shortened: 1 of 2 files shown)


phillipbarron (approved):
👍

twrichards:
✅  TESTED in `CODE`

PR #210, opened 2022-11-08
display unread from team pinboards

twrichards:
Co-authored-by: @aracho1 

Now that we have the `MY TEAMS' PINBOARDS` list on the `SelectPinboard` view (thanks to #206), we want to be able to surface any unread group mentions which are unclaimed or informational via our existing unread flag mechanism (i.e. a red blob on the Floaty and a black blob on the relevant pinboard button on the `SelectPinboard` view).

- tweak `getGroupPinboardIds` query to include `hasUnread` property (for unclaimed or informational group mentions only) - unread being calculated by doing an inner query on the `LastItemSeenByUser` table
- poll the above every 5s, regardless of whether pinboard is expanded
- utilise those `hasUnread` values when processing the `getGroupPinboardIds` response, calling the existing `setUnreadFlag` for each pinboard
- change the sort function for the `MY TEAMS' PINBOARDS` list to place unread at the top

ALSO INCLUDES changes from https://github.com/guardian/pinboard/pull/211 (due to an accidental/premature merge of chained PR)

### DIFF
--- a/cdk/test/__snapshots__/stack.test.ts.snap
+++ b/cdk/test/__snapshots__/stack.test.ts.snap
@@ -1217,6 +1217,7 @@ type PinboardIdWithClaimCounts {
   othersClaimedCount: Int!
   notClaimableCount: Int!
   latestGroupMentionItemId: String!
+  hasUnread: Boolean!
 }
 ",
       },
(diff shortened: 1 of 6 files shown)


phillipbarron (approved):
LGTM 👍

PR #212, opened 2022-11-08
load `Guardian Agate Sans` font (if not already loaded by host platform)

twrichards:
Co-authored-by: @aracho1 

https://trello.com/c/am4GCw1H/946-load-agate-fonts-in-pinboard

During a recent demo we noticed pinboard (within grid) was displaying with basic sans font (rather than `Guardian Agate Sans`) because Pinboard previously relied upon the host platform to load the Agate font, but since https://github.com/guardian/grid/pull/3893 we cannot rely on this.

This PR moves the approach used for the local testing playground (loading the Agate fonts from `interactive.guim.co.uk`) to the top level of the React tree (in `app.tsx`). Note that the `@font-face` declarations need to be made outside of the shadow DOM due to browser restrictions.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -6,7 +6,6 @@ set -e
 yarn lint-staged
 
 if (git diff --name-only --cached | grep -e 'cdk' -e 'shared/graphql'); then
-    yarn graphql-refresh
     yarn workspace cdk update-snapshot
     git add cdk/test/__snapshots__/
 fi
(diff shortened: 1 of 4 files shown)


twrichards:
not great, this approach loads them every time in composer, despite composer loading them already - boo.

also grid needs the `font-src` bit of the Content-Security-Policy updating (probably via a new config option - _sigh_)...
https://github.com/guardian/grid/blob/6634ff0f0443c78fdad36e09e9c633728953eba5/kahuna/app/KahunaComponents.scala#L59

phillipbarron (approved):
LGTM 👍

twrichards:
I explored bundling the `@font-face` CSS separately to keep it out the bundle but it's only a 1KB difference, so not worth the complexity of dynamic imports, webpack config etc.

PR #216, opened 2022-11-16
'ASCII fold' the `searchText` when searching workflow, such that we match diacritics

twrichards:
... probably needs some work doing in workflow too, to actually work.


twrichards:
closing this as this really needs to happen in workflow DB query, but that has performance implications and requires an extension to be installed - https://stackoverflow.com/a/13597082 - so probs not worth it

PR #219, opened 2022-12-01
column in workflow

twrichards:
Co-authored-by: @aracho1 

https://trello.com/c/s5n4JW0z/176-add-pinboard-visibility-in-a-new-workflow-column

**RELATED: https://github.com/guardian/workflow-frontend/pull/385**

workflow_pinboard

This PR makes pinboard populate the new [`Pinboard 📌` column in workflow](https://github.com/guardian/workflow-frontend/pull/385). When the hostname starts with `workflow.` pinboard runs in '_inline mode_' - so instead of adding the `Floaty` and `Panel` to the page it adds a new component `InlineMode` (`inline/inlineMode.tsx`) which takes in one prop `workflowPinboardElements` (a list of DOM elements with class `content-list-item__field--pinboard` - see https://github.com/guardian/workflow-frontend/pull/385), found using the existing MutationObserver (albeit with some tweaks to fire on attribute changes, so we can hear about 'working title' and 'headline' changes from workflow, which come via data attributes). 

With this list of DOM elements (for each cell of the `Pinboard 📌` column) we extract the pinboard ID (workflow id) to build a new query called `getItemCounts` (which we also poll on a 5s frequency), this runs a DB query, joining a few tables, to return an `unreadCount` and `totalCount` for each pinboard Id. 

With these counts per pinboard we we use [React Portals](https://reactjs.org/docs/portals.html) (like elsewhere in the codebase) to take over the DOM elements (each cell of the `Pinboard 📌` column) and render a new component `InlineModePinboardToggle` (`inline/inlineModePinboardToggle.tsx`) to display these counts and make the cell clickable to load the pinboard...

...via the state (for the selected pinboard) held in `InlineMode` (`inline/inlineMode.tsx`) and triggers the mounting of `InlineModePanel`(`inline/inlineModePanel.tsx`) into the `pinboard-area` element reserved by workflow (see https://github.com/guardian/workflow-frontend/pull/385). This panel brings together the existing `Navigation` component (albeit with a few little refactors) and an instance of the existing `Pinboard` component. It does extract the `data-working-title` and `data-headline` to flow into the Navigation.

We also take over the pinboard column heading (`.content-list-head__heading--pinboard`) which is empty and add a custom heading (`InlineModeWorkflowColumnHeading` in `inline/inlineModeWorkflowColumnHeading.tsx`) with some nice SVG icons...
image

Lastly, when a row with an expanded/selected pinboard is scrolled out of view it is now _sticky_, and the subsequent group headings are offset to produce the experience below...
workflow_pinboard_sticky


aracho1 (approved):
🚀 LGTM

PR #220, opened 2022-12-06
use `ResizeObserver` for the `scrollToBottomIfApplicable` behaviour for greater reliability

twrichards:
Despite efforts in https://github.com/guardian/pinboard/pull/169 there were still occasions where it layout shift in the items in scrollable area would mean we weren't quite scrolled to the bottom, so now as soon as we have content in the scrollable area we register a [`ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) to listen for changes in height of the scrollable content (i.e. all the `ItemDisplays`) and if we do the previous `scrollToBottomIfApplicable` behaviour (i.e. if we're currently scrolled to bottom then keep it that way by scrolling to the bottom, e.g. new message comes in or one of the `GridDynamicSearchDisplay` reloads).

This means we could phase out passing the `scrollToBottomIfApplicable` function around, thereby simplifying some of the child components 🎉

### DIFF
--- a/client/src/grid/gridDynamicSearchDisplay.tsx
+++ b/client/src/grid/gridDynamicSearchDisplay.tsx
@@ -14,10 +14,6 @@ import { FormattedDateTime } from "../formattedDateTime";
 import { GridBadge } from "./gridBadges";
 import { SvgReload } from "@guardian/source-react-components";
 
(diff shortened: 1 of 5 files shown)


phillipbarron (approved):
nice 👍

PR #221, opened 2022-12-06
add a new 'dormant overlay' when the browser tab doesn't have focus

twrichards:
During one of @CDicksonG's recent demos I noticed he struggled to demo the 'unread' flags because he had other tabs open which were immediately marking the new messages as read. 

This PR detects when the tab is in focus (using `blur`, `focus` and `visibilitychange` event listeners) and...

- prevents auto-scrolling to the bottom when new messages come in
- and prevents new messages being marked as read (because of the above scroll prevention)
- adds an overlay which explains ` When pinboard is not in the active browser tab, it will still receive messages, but not mark them as seen.`

dormant_mode


twrichards:
After a thought from @alinaboghiu - I've reconsidered the overlay part of this and decided it's a distraction (not to mention the nasty interplay with the `Find` functionality in browsers), so I've removed it.

> - adds an overlay which explains ` When Pinboard is not in the active browser tab, it will still receive messages, but not mark them as seen.`
> 
> dormant_mode

phillipbarron (approved):
LGTM :+1

client/src/globalState.tsx
```diff
@@ -26,6 +26,9 @@ import { UserLookup } from "./types/UserLookup";
 
 const LOCAL_STORAGE_KEY_EXPLICIT_POSITION = "pinboard-explicit-position";
 
+const getHasBrowserFocus = () =>
```
alinaboghiu, line 29: loving the naming refinement too

alinaboghiu (approved):

PR #223, opened 2022-12-08
attempt to `focus` relevant browser tab when clicking a desktop/push notification

twrichards:
_For more information about notifications mechanism see https://github.com/guardian/pinboard/tree/main/notifications-lambda#notifications-lambda_

When experimenting in `CODE` users noticed that clicking a desktop notification when pinboard was open in a non-active tab did not bring the tab to the foreground. This PR looks to address that by calling [`WindowClient.focus()`](https://developer.mozilla.org/en-US/docs/Web/API/WindowClient/focus) in the service worker.

With multiple tabs open which have pinboard loaded, the notification click needs to focus the correct tab ideally. So now, the hidden iFrame (which facilitates cross-domain communication between the pinboard running in the host application and the service worker running on the pinboard domain) receives the selected pinboard id via a query param, which allows the service worker to know which `WindowClient` to focus.

notification_highlight

### DIFF
--- a/client/src/app.tsx
+++ b/client/src/app.tsx
@@ -267,24 +267,32 @@ export const PinBoardApp = ({ apolloClient, userEmail }: PinBoardAppProps) => {
 
   const showDesktopNotification = (item?: Item) => {
     if (item && item.userEmail !== userEmail) {
-      serviceWorkerIFrameRef.current?.contentWindow?.postMessage(
-        {
-          item: {
-            ...item,
-            payload: item.payload && JSON.parse(item.payload),
-          } as ItemWithParsedPayload,
-        },
-        desktopNotificationsPreferencesUrl
+      setTimeout(
+        () =>
+          serviceWorkerIFrameRef.current?.contentWindow?.postMessage(
+            {
+              item: {
+                ...item,
+                payload: item.payload && JSON.parse(item.payload),
+              } as ItemWithParsedPayload,
+            },
+            "*"
+          ),
+        500
       );
     }
   };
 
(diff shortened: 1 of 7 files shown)


client/src/app.tsx
```diff
@@ -336,53 +336,53 @@ export const PinBoardApp = ({ apolloClient, userEmail }: PinBoardAppProps) => {
   return (
     <TelemetryContext.Provider value={sendTelemetryEvent}>
       <ApolloProvider client={apolloClient}>
-        <Global styles={agateFontFaceIfApplicable} />
-        <HiddenIFrameForServiceWorker iFrameRef={serviceWorkerIFrameRef} />
-        <root.div
-          onDragOver={(event) =>
-            isGridDragEvent(event) && event.preventDefault()
+        <GlobalStateProvider
```
twrichards, line 339: changes in this file seem a lot but its just moving the `GlobalStateProvider` slightly further up the hierarchy so the `HiddenIFrameForServiceWorker` has access to the global state (see that changes in that file).

aracho1 (approved):
Tested on CODE, works great ✨

aracho1:
Just noticed this console error which we might want to address
image

twrichards:
> Just noticed this console error image
> 
> a quick google search seems to suggest this will log when the iframe has not yet loaded but since service worker will keep posting messages and everything seems to work fine, should be fine to ignore?

Excellent spot @aracho1 - thank you - and also thank you for googling the likely reason - whilst it's good it still works, I'm going to spend a short while seeing if I can delay the post message until the iFrame has loaded.

PR #225, opened 2022-12-13
fix `userDisplayName` on `ClaimableItem`

twrichards:
Noticed that `ClaimableItem` component was not rendering a the user name in a friendly way, e.g. 
image
(notice the email address instead of first name and last name).

This was because `ClaimableItem` was wrapped in a `useMemo` (for improved performance) which didn't have the `userDisplayName` in its dependencies list and since `userLookup` (from which `userDisplayName` is derived) is populated asynchronously, it never got updated and so displayed the fallback (which is the email address).

**This PR ensures the `userDisplayName` actually flows into `ClaimableItem` when its updated asynchronously.**

### DIFF
--- a/client/src/itemDisplay.tsx
+++ b/client/src/itemDisplay.tsx
@@ -207,7 +207,7 @@ export const ItemDisplay = ({
               }
             />
           ),
-          [maybeClaimedBy]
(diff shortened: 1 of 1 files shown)


aracho1 (approved):
Good spot

PR #230, opened 2022-12-22
Edit & Delete items

twrichards:
https://trello.com/c/tsDzYGvA/997-add-edit-and-delete-functions-to-pinboard-items-on-hover

This is something we've long considered to be necessary at some point but when demoing with Editorial teams it has become apparent this is a rather crucial feature (and is needed before wide rollout).

edit delete

### Hover Menu
We introduce a new hover menu on each item, which contains...
- a pencil icon to enter edit mode
- and a bin icon (new to the repo) to display a delete confirmation modal

... and displays only if...

- for your own messages
- if you're not currently editing 
- the item is not already deleted
- the item is not a 'claim' of another item
 
image

_This menu also provides a future home for features such as 'reply', 'starred messages' and more._

### Delete
- Add `deletedAt` timestamp column to the `Item` table
- Add `deleteItem` GraphQL mutation, which simply takes the itemId as an argument
- The `deleteItem` handler in `database-bridge-lambda` clears the message and payload fields (in case deleting because sensitive) and set the new `deletedAt` column to the current time - this way we can still pull aggregated data (such as unique users).
- Filtered out deleted items from `getGroupPinboardIds` (used for the `MY TEAMS' PINBOARDS` list)  and `getItemCounts` (used to populate the workflow column).
- Clicking image in the new hover menu presents a modal to confirm deletion before invoking the new `deleteItem` GraphQL mutation, it presents the item message in a height limited mini scrollable pane within the modal to ensure user knows exactly what they're deleting.

### Edit
- Add `editHistory ` timestamp array column to the `Item` table
- Add `editItem` GraphQL mutation, which takes the itemId and an `EditItemInput` (new type with `message`, `payload` and `type` fields) as arguments
- The `editItem` handler in `database-bridge-lambda` updates `message`, `payload` and `type` fields accordingly, but also adds the current time to the new `editHistory` timestamp array column.
- Clicking image in the new hover menu will place the pinboard into an edit mode (controlled at the pinboard level with new state `maybeEditingItemId`) - this changes the `ItemDisplay` to a new `EditItem` component, which renders the `ItemInputBox` (renamed from `CreateItemInputBox` - with some slight refactors, including suppressing the ability to mention people) plus a two buttons, one to cancel and one to actually perform the new `editItem` GraphQL mutation.
Note that any `payloadToBeSent` that's currently set, will be stored off to a temporary bit of state to ensure it's not shown in the background and will restored once editing is complete/cancelled - so user can select a different payload if thats what they want to do in their edit operation.

### Edit & Delete 
- Renamed the `onCreateItem` subscription to be `onMutateItem` and added `editItem` and `deleteItem` alongside `createItem` in the list of mutations its firing on - since they all return the new/updated `Item` and the client already has the logic to update the `itemMap` etc to update the UI and so on.

### Telemetry
We introduce 4 new event types...
- `DELETE_ITEM`
- `CANCEL_DELETE_ITEM`
- `UPDATE_ITEM`
- `CANCEL_UPDATE_ITEM` 

...which you can see visible in Grafana...
image

### PR actions
Run `yarn database-setup`, selecting `add editHistory and deletedAt columns to Item table` against...
- [x] CODE
- [x] PROD


pradasa:
Hi @twrichards 
The UX of this works great! Thanks for adding it. Would it be okay to make a tiny adjustment on the position of the update/cancel buttons, please?

Would it be possible to:
- align them to the right (where update is at the moment) 
- place update first, then cancel (opposite order to the current one)
- keep the same styling you have for the buttons
- leave 4px spacing between buttons

Cheers!

client/src/itemDisplay.tsx
```diff
@@ -145,69 +137,101 @@ export const ItemDisplay = ({
...
             margin-bottom: 2px;
           `}
         >
           <FormattedDateTime timestamp={dateInMillisecs} />
+          {isEdited && (
+            <span>
+              &nbsp;-&nbsp;<em>Edited</em>
+            </span>
+          )}
+          {!isDeleted &&
```
phillipbarron, line 152: there are a lot of conditions here, maybe extract to a meaningful const like `isEditableItem` or similar

client/src/itemInputBox.tsx
```diff
@@ -156,12 +156,19 @@ export const CreateItemInputBox = ({
     []
   );
 
+  const resizeTextArea = (target: HTMLTextAreaElement) => {
+    target.style.height = "0";
+    // Chrome will sometimes show a scrollbar at the exact scrollHeight, so give it a .1px extra as a nudge not to add one
+    target.style.height = `${target.scrollHeight + 0.1}px`;
```
phillipbarron, line 162: 😞

client/src/itemDisplay.tsx
twrichards: good suggestion, thanks - done in https://github.com/guardian/pinboard/pull/230/commits/42cd458c05a1a9f2f540c034a220d9747d724391

client/src/itemDisplay.tsx
```diff
@@ -91,7 +91,8 @@ export const ItemDisplay = ({
     !isDeleted &&
     !maybeEditingItemId &&
     item.userEmail === userEmail &&
-    item.type !== "claim";
+    item.type !== "claim" &&
+    !item.claimedByEmail;
```
phillipbarron, line 95: ❤️

phillipbarron (approved):
This is brilliant, Tom

paperboyo:
💎

PR #231, opened 2022-12-22
catch auth failures and prevent further requests (rather than spamming Sentry)

twrichards:
In https://github.com/guardian/pinboard/pull/224 we fixed `auth-lambda` from erroring on expired tokens (typically from polling in stale browser tabs) and at that time the Sentry quota was exhausted so we missed the fact that the client will still report this same volume of failures but now `401`s rather than `500`s essentially.

This PR catches auth failures and sets some global state, which is used to **swallow subsequent requests** and also to present the existing error overlay (although refactored into its own component) but now not just within each `Pinboard` but at the `Panel` and `InlineModePanel` level (also ensuring it displays correctly and doesn't burst out of its container).


phillipbarron (approved):
LGTM 👍

PR #235, opened 2023-02-13
adopt guCDK (first pass - all constructs remain vanilla CDK)

twrichards:
https://trello.com/c/OlfYW4Xf/804-adopt-gucdk-in-pinboard

> **Warning**
> https://github.com/guardian/pinboard/pull/235/commits/41b298a466420f4c788bf5be6051d373af706685 MUST be deployed/cloudformed FIRST and then this PR can be merged/deployed - this is to workaround CloudFormation constraints/bugs about modifying custom named resources.
> ### Applied to...
>- [x] CODE
>- [x] PROD

When we first wrote the infra for pinboard, the [`@guardian/cdk`](https://github.com/guardian/cdk) project was still in development and not ready for adoption, so instead we use vanilla CDK - 2 years on, we're finally getting round to rectifying that.

This PR is the **initial first pass**, that adds the  [`@guardian/cdk`](https://github.com/guardian/cdk) lib and refactors to extend **`GuStack`** rather than just `Stack` (including splitting the resulting template to be per stage, to adopt guCDK convention), this alone results in a huge CFN change set (albeit mainly tagging etc.)...
image
... note that the 'replace' on the RDS instance is a false positive in the change set, @aracho1 and I tested this with @akash1810 and @JuliaBrigitte in CODE.

This PR does NOT include refactoring any resource definitions to make use of guCDK patterns/constructs (abstractions) primarily because this PR is already huge and so don't want to make it any harder to review, but also so that this and subsequent PRs (to refactor each resource) are easier to review as well as being useful examples for other projects/teams.


twrichards:
As per description, this as has already been applied to both `CODE` and `PROD`.

cdk/lib/stack.test.ts
```diff
@@ -1,10 +1,14 @@
...
 
 describe("PinBoardStack's", () => {
   it("generated CloudFormation matches the snapshot", () => {
     const app = new App();
-    const stack = new PinBoardStack(app, "PinBoardStack");
+    const stack = new PinBoardStack(app, "PinBoardStack-TEST", {
+      domainName: "pinboard.test.dev-gutools.co.uk",
+      stack: "workflow",
+      stage: "TEST",
+    });
```
aracho1, line 11: Just for my understanding, what does this do?

aracho1 (approved):
LGTM 👍

cdk/lib/stack.test.ts
twrichards: This is the file which triggers the snapshot tests. And since the stage specific variables are passed as props to `PinboardStack` we supply test related ones in the test.

Snapshot tests allow us to see a more verbose/representative view of the infrastructure changes we're making, compared to the abstraction that is CDK, and further abstraction that is guCDK. E.g. one line of CDK can result in many lines of cloudformation. So we make the snapshot also part of VCS so in PR code review we can see not only the diff in CDK code but also the resulting diff in generated cloudformation.

PR #236, opened 2023-02-14
cleanup some unused variables

twrichards:
ronseal


aracho1 (approved):
💯 + extra points for the gif

PR #238, opened 2023-02-27
convert database jump host to use `GuAutoScalingGroup`

twrichards:
Following on from #235, the database jump host ASG is the first thing we're migrating from vanilla CDK to a guCDK construct. This also includes...
- using `GuAlarm` for the accompanying custom alarm
- refactoring the 'user data' script to be a deployable artifact* `startup.sh` which takes arguments (rather than string replace which was a bit gross tbf)

\* worth noting that the changes to `artifact.json` and `riff-raff.yaml` are now needed, but will hopefully be shortlived until we adopt guCDK's [`riff-raff.yaml` generation](https://github.com/guardian/cdk/tree/8dfedd79964322259bc14e3e1d23da728ae33b60/src/experimental/riff-raff-yaml-file) - our zero sized ASG might prove tricky there but we can always raise a guardian/cdk PR if needs be

## Tested in `CODE` ✅
Following some naming tweaks in `getDatabaseJumpHost.ts` (since logical IDs have changed), we can connect to an instance and then on to the DB and execute SQL 🎉

### DIFF
--- a/artifact.json
+++ b/artifact.json
@@ -13,6 +13,11 @@
       "compress": false
     },
     {
+      "action": "pinboard-db-jump-host-asg-artifact",
+      "path": "cdk/startup.sh",
+      "compress": false
+    },
+    {
       "action": "pinboard-bootstrapping-lambda-api",
       "path": "dist",
       "compress": "zip"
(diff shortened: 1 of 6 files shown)


twrichards:
Whilst overall this is an increase in LoC (which is down to the snapshot, because guCDK constructs have more implicit stuff, typically for best practice configuration etc.) the `stack.ts` file is reduction...
image

twrichards:
### problems
- [ ] self-terminating thing looks to be broken, likely permissions issue
- [ ] the alarm that the instance had been running a for too long also didn't fire

twrichards:
applied to this to `PROD` manually as it doesn't like replacing an ASG with custom name, but that's now done so merging...

PR #240, opened 2023-03-08
turn on OK actions for the `DatabaseJumpHostOverrunningAlarm`

twrichards:
In #238 it looks like we lost the OK actions on the `DatabaseJumpHostOverrunningAlarm` which means it stays in 'error' state in PagerDuty, gChat etc.


phillipbarron (approved):
LGTM 👍

PR #241, opened 2023-03-08
add `archiver-lambda` to regularly clean up after workflow items are deleted

twrichards:
In #117 we added a tactical feature to workaround the fact that workflow items disappear (see [guardian/workflow/archiver](https://github.com/guardian/workflow/tree/main/archiver) and so we can't retrieve the working title etc, so we displayed `NOT FOUND` in the UI - which was useful at the time so we could review the content of pinboard during live blog testing for example.

https://trello.com/c/kISy1TpU/996-archive-pinboard-items-when-associated-workflow-stub-is-purged

Longer term we'd like the pinboards associated with deleted workflow items to just drop away, to keep things relevant for users. This PR aims to address that, by...

- introduces a new `isArchived` column on the `Item` table (with an index)
- filters out items with `isArchived` = `true` from relevant queries
   - `listItems` - the query performed to populate all the messages/items in an individual pinboard
   - `getGroupPinboardIds` - the query performed to populate the `MY TEAMS' PINBOARDS`
   - `getItemCounts` - the query performed to get counts for the workflow column (see #219)
- adds a new `archiver-lambda` (making use of `GuScheduledLambda` - very easy thanks to #235) which runs every 6 hours and does the following...

   - invokes `workflow-bridge-lambda` to get active workflow IDs, including trashed but not yet actually deleted (this required a small refactor)
   - for items with `pinboardId` NOT in the list of active workflow IDs
      - sets `isArchived` to `true`
      - clear the `message` and `payload` for storage reasons [FYI @blishen @hoyla ]
   - removes any archived pinboards from the `manuallyOpenedPinboardIds` field of relevant users

   ... details of the above is all logged, such that we could restore if needs be (albeit with an ad-hoc script).

See updated architecture diagram for how this fits in...
Pinboard Architecture (1)


twrichards:
TESTED on `CODE` ✅

PR #248, opened 2023-03-22
Upgrade `prettier`

twrichards:
because...
A) it's best practice 
B) following #247 , we want to use nice new features of TypeScript such as `satisfies` keyword (see https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/#satisfies) which `prettier` didn't like

... this upgrade changes prettier's behaviour slightly so **re-formats a number of files as part of this diff**


aracho1 (approved):
🧼💯

PR #251, opened 2023-04-12
upgrade node from 16 to 18 AND exclude `aws-sdk` and `@aws-sdk` from lambda ZIPs

twrichards:
This PR upgrades to latest Node [which is supported in Lambda env].

Importantly ALSO we now exclude `aws-sdk` and `@aws-sdk` from lambda ZIPs, since according to https://docs.aws.amazon.com/lambda/latest/dg/lambda-nodejs.html v3 of AWS SDK is provided at runtime (we've been bundling `@aws-sdk` deps in with the lambdas since #249)  and it makes more sense for Pinboard to keep lambdas as small as possible.

PLUS since Node 18 now provides `fetch` natively, we've been able to remove `node-fetch` dependency 🎉  from...

- `workflow-bridge-lambda`
- `grid-bridge-lambda`
- `users-refresher-lambda`

... NOTE however that the `@types/node-fetch` dev dependency remains as does the import of `node-fetch` otherwise we get `TS2304: Cannot find name 'fetch' ` (see issue https://github.com/DefinitelyTyped/DefinitelyTyped/issues/60924).


twrichards:
I've tested this in `CODE` 🎉  - will update the description on #249 to reference this once this is merged.

aracho1 (approved):
🚀

PR #254, opened 2023-04-17
Add telemetry to Tour

aracho1:
## What does this change?
Set up a basic client-side telemetry when users:
- start the tour
- select a step from the initial Contents step (to track most 'popular' step)
- try sending interactive messaging in the Messaging step
- close the tour (before finishing)
- finish the tour in the last step

The events are captured as one telemetry type (i.e.`PINBOARD_TELEMETRY_TYPE.INTERACTIVE_TOUR`) but as separate tags

## How to test
Tested on CODE ✅

Example Grafana panel 
image
https://metrics.gutools.co.uk/d/b-tyvbs7z/pinboard?orgId=1&from=now-30m&to=now&refresh=1m&var-Stage=CODE&var-Interval=$__auto_interval_Interval&viewPanel=37

Interacting with the tour should send telemetry data in the Network tab 'event' e.g.
image

Grafana dashboard should log user interactions with the tour

Tested on CODE

### DIFF
--- a/client/src/sendMessageArea.tsx
+++ b/client/src/sendMessageArea.tsx
@@ -120,17 +120,8 @@ export const SendMessageArea = ({
 
   const sendItem = () =>
     confirmClaimable(verifiedGroupMentionShorthands?.length > 0).then(
-      (claimable) =>
-        (tourProgress.isRunning
-          ? (sendTelemetryEvent?.(PINBOARD_TELEMETRY_TYPE.INTERACTIVE_TOUR, {
-              tourEvent: "messaging",
-            }),
-            tourProgress.sendItem(() => {
-              setMessage("");
-              clearPayloadToBeSent();
-              setUnverifiedMentions([]);
-            }))
-          : _sendItem)({
+      (claimable) => {
+        const messageItem = {
           variables: {
             input: {
               type: payloadToBeSent?.type || "message-only",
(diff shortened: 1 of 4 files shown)


client/src/sendMessageArea.tsx
```diff
@@ -122,11 +122,14 @@ export const SendMessageArea = ({
...
-          ? tourProgress.sendItem(() => {
+          ? (sendTelemetryEvent?.(PINBOARD_TELEMETRY_TYPE.INTERACTIVE_TOUR, {
+              tourEvent: "messaging",
+            }),
+            tourProgress.sendItem(() => {
               setMessage("");
               clearPayloadToBeSent();
               setUnverifiedMentions([]);
-            })
+            }))
```
twrichards, line 132: this looks a little odd, I'd have thought the telemetry event would be called in that callback which clears the message etc.

twrichards (approved):

client/src/sendMessageArea.tsx
aracho1: Seemed like the only way to keep the ternary (to avoid having to repeat lines 135-143 in this file) while calling multiple functions ([reference](https://github.com/guardian/pinboard/pull/254/files/91bff7bb8dd2dfc04859b1c367ddedbddcb6bb8d#diff-a2a48a4733d8d8a4c55d4e1013b725684b24485ca744531d25abf49f4e3a13b6R136-R143)), but maybe I should just use if/else for more clarity.

PR #257, opened 2023-04-18
detect grid links on paste and convert to proper payload

twrichards:
https://trello.com/c/jvhO7GKf/619-pinboard-turn-grid-links-into-proper-payloads

We observed in PROD that users are still pasting grid links (probably muscle memory from doing so with email, chat etc.)...
image

So now we detect on paste (rather than on send) - if the pasted text is a grid link we call a NEW `asGridPayload` GraphQL Query, handled by `grid-bridge-lambda` which based on the URL builds the payload (calling out to grid API if needs be) and type (`grid-search`, `grid-original` or `grid-crop`) and calls the `setPayloadToBeSent` (just like `Add to 📌` and drag'n'drop does) before finally removing the pasted grid link from the message). Note we've retained the telemetry event.

paste_grid_links


phillipbarron (approved):
Tested locally, works beautifully 👍

PR #259, opened 2023-04-21
add `email-lambda` to email users about missed messages etc.

twrichards:
Run `yarn database-database` setup, selecting the stage and then the last option (to effectively run `018-AddIsEmailEvaluatedColumnToItemTable.sql`) for
- [x] **`CODE`**
- [x] **`PROD`**

---

https://trello.com/c/DofOq5iy/1333-spike-add-email-lambda-to-pinboard-to-send-emails-for-missed-messages

If users are not in the tools and don't have push notifications turned on there's a good chance they might miss an important message. This PR adds a new `isEmailEvaluated` boolean column to the DB and introduces a new lambda which runs on a schedule (every 5mins) it...

1. ~queries the DB for items with mentions (individual only for now, group mentions will come in a follow-up PR) which have remained unread for more than an hour [by the mentioned user(s)] which hasn't had it's `isEmailEvaluated` set to true.~ (see #266)
1. queries the DB for items where `Central Production` is group mentioned and hasn't already had it's `isEmailEvaluated` set to true.
2. collates all items into per-user data structure, and then per-pinboard within that
3. looks-up the working titles and headlines from workflow (by invoking `workflow-bridge-lambda`) 
4. loops through the users, sending them an email (using [AWS Simple Email Service (SES)](https://aws.amazon.com/ses/)) from the relevant pinboard domain - `mentions@pinboard.code.dev-gutools.co.uk` or `mentions@pinboard.gutools.co.uk` 
   - These domains are registered as 'SES verified identities' and DNS validated automatically via CDK code 🎉 . 
   - The email is rendered using `preact` (like the client) but using `style` attribute rather than emotion, for best compatibility with email clients.
5. sets `isEmailEvaluated` to true for the items it has emailed about 

image

(note, only CODE emails are flagged as `External` - because `gutools.co.uk` is a valid alias in gSuite, but `code.dev-gutools.co.uk` is not)

### Architecture Changes
Pinboard Architecture (diff)


pradasa:
Hi Tom, 

I had an initial pass at the UI. 
Screenshot 2023-05-22 at 15 15 19

It would be nice to pull who has sent the message if possible (like Trello and Slack do). 
In regards to the UI, having text on neutral backgrounds will help with readability. 
We can change the yellow box above for a top yellow rule and a grey background. It's a good shot to make sure it's easy to differentiate that they belong to different pinboards. 

I like the idea of the links going directly to each message, will make things less confusing. 
I'm going to look at the image thumbnail tomorrow AM

pradasa:
Just to gather everything discussed over chat here:

Language-wise
1)We want to change the email content title to "you might have missed..." It's shorter, a bit clearer and we can be playful with this one. 
2) We're still discussing the subject of the email

Content-wise
1) fonts should always be system-font, not Agate or Guardian fonts. (I'm using Inter on the mock ups, but it should be replaced as required). 
2) We want to add a one liner explaining why users are getting these emails. 
3) The font size for the WT/HL footnote needs to increase so it's not lost at the bottom of the email. We don't want to place it at the top and make it too prominant. 
4) Whenever an image is added, we can use the same style of images in "chat view" except we don't add the metadata that we display there (ie. number of images, refresh icon, label)

We also discussed the idea of adding timestamps for context on missed mentions but decided against them because of time zones and how they might complicate timestamp implementation.

twrichards:
Tested in `CODE`
image

Fweddi (approved):
Pair reviewed in person. LGTM - we are releasing the email functionality to CP first, for any group mention of them. In future we will also roll out any personal mentions that have not been seen within an hour.

PR #260, opened 2023-05-03
trigger interactive tour on first visit to Pinboard

aracho1:
Co-authored-by: @twrichards 

## What does this change?

This PR adds an additional `visitedTourSteps` column in the User table to track the steps each user has navigated to in the interactive tour. If a user has never used the tour (i.e. if the column is null), the tour starts by default when they expand the floaty. If the user dismisses the tour (by closing the panel), this will be saved in the column (as "DISMISSED") so that they are not shown it again.

## How to test
Tested on CODE:
- The database saves each tour step correctly 
- Tour starts automatically if the user has never used the tour
- Tour does NOT start again when the user dismisses the tour with or without refreshing the page

Tour starts by default for new users but is still easy to dismiss

Tested on CODE

### DIFF
--- a/client/src/globalState.tsx
+++ b/client/src/globalState.tsx
@@ -60,6 +60,7 @@ interface GlobalStateContextShape {
   selectedPinboardId: string | null | undefined;
   clearSelectedPinboard: () => void;
 
+  hasEverUsedTour: boolean | undefined;
   addCompletedTourStep: (tourStepId: string) => void;
 
   showNotification: (item: Item) => void;
@@ -121,6 +122,7 @@ interface GlobalStateProviderProps {
   showNotification: (item: Item) => void;
   clearDesktopNotificationsForPinboardId: (pinboardId: string) => void;
   presetUnreadNotificationCount: number | undefined;
+  hasEverUsedTour: boolean | undefined;
 }
 export const GlobalStateProvider: React.FC<GlobalStateProviderProps> = ({
   hasApolloAuthError,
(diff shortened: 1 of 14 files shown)


shared/graphql/schema.graphql
```diff
@@ -31,6 +31,7 @@ type Mutation {
   setWebPushSubscriptionForUser(webPushSubscription: AWSJSON): MyUser
   addManuallyOpenedPinboardIds(pinboardId: String!, maybeEmailOverride: String): MyUser
   removeManuallyOpenedPinboardIds(pinboardIdToClose: String!): MyUser
+  addVisitedTourStep(tourStepId: String!): MyUser
```
twrichards, line 34: probably not worth changing now, but could be...
```suggestion
  visitTourStep(tourStepId: String!): MyUser
```
or
```suggestion
  visitedTourStep(tourStepId: String!): MyUser
```

client/src/tour/tourState.tsx
```diff
@@ -239,6 +255,7 @@ export const TourStateProvider: React.FC = ({ children }) => {
             nextStepIndex,
           ]);
           continueTourTo(nextStepIndex);
+          addVisitedTourStep(tourStepIDs[index]); // tracks in the database TODO - rename to addVisitedTourStep
```
twrichards, line 258:
```suggestion
          addVisitedTourStep(tourStepIDs[index]); // tracks in the database
```

client/src/globalState.tsx
```diff
@@ -517,6 +524,9 @@ export const GlobalStateProvider: React.FC<GlobalStateProviderProps> = ({
     selectedPinboardId,
     clearSelectedPinboard,
 
+    hasEverUsedTour,
+    addVisitedTourStep: addVisitedTourStep,
```
twrichards, line 528:
```suggestion
    addVisitedTourStep,
```

twrichards (approved):
amazing work!

PR #263, opened 2023-05-17
open pinboards in new composer tab always when `Shift` key is pressed in `SelectPinboard`

twrichards:
User request to be able to open the composer piece directly from the search results (e.g. to check its the right composer piece when in grid for example, without clicking into it and then having to click the composer button, meanwhile it being added it to your 'My Pinboards').

shift_and_click

### DIFF
--- a/client/src/selectPinboard.tsx
+++ b/client/src/selectPinboard.tsx
@@ -30,6 +30,7 @@ import { NotTrackedInWorkflow } from "./notTrackedInWorkflow";
 import { Feedback } from "./feedback";
 import { useTourProgress, useTourStepRef } from "./tour/tourState";
 import { demoPinboardData } from "./tour/tourConstants";
+import { ComposerIcon } from "./navigation/icon";
 
 const textMarginCss: CSSObject = {
   margin: `${space["1"]}px ${space["2"]}px`,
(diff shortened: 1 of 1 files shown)


simonbyford (commented):
Looks good!

I'm getting some buggy behaviour with the shift detection. To reproduce:

- go [here](https://composer.code.dev-gutools.co.uk/?expandPinboard=true) 
- move cursor into the pinboard area
- stop moving the cursor
- press shift and observe nothing happens

If you press shift while moving the cursor, the problem is fixed.

However this is a bit intermittent for me which makes me wonder if it's just my browser. I'm on chrome 113.0.5672.126

client/src/selectPinboard.tsx
```diff
@@ -459,16 +481,21 @@ export const SelectPinboard = ({
           </div>
         </div>
         {loading && searchText && <Text>Searching...</Text>}
+        {data && searchText && (
+          <Text>
+            {unopenedPinboards.length > 0
+              ? "Shift + click opens in Composer"
```
simonbyford, line 487: Would it be worth including this note under the "MY TEAMS' PINBOARDS" header as well as the "SEARCH" header? As the functionality seems to apply to both sections

twrichards: astute observation. I did consider that but its only directly useful for the search results, and the fact the same behaviour applies to the 'my pinboards' and 'my teams pinboards' is because they share a component, and I'm not sure it warrants passing in as a prop whether this new shift behaviour should apply

simonbyford: Yup, sounds reasonable!

twrichards:
> Looks good!
> 
> I'm getting some buggy behaviour with the shift detection. To reproduce:
> 
> * go [here](https://composer.code.dev-gutools.co.uk/?expandPinboard=true)
> * move cursor into the pinboard area
> * stop moving the cursor
> * press shift and observe nothing happens
> 
> If you press shift while moving the cursor, the problem is fixed.
> 
> However this is a bit intermittent for me which makes me wonder if it's just my browser. I'm on chrome 113.0.5672.126

Thanks @simonbyford - I wasn't super happy with the implementation and knowing its noticeable for others, I've re-written as window level event listeners (added and removed via a useEffect)

simonbyford (approved):
Looking perfect now! 🚢

PR #264, opened 2023-05-19
add a spinner during pasted grid link resolution and block sending

twrichards:
In #257 we began detecting grid links being pasted and convert them to real payloads. However this was done silently in the background, and as a result we observed in the wild some links still being sent as is, the theory being that the users hit send before the payload had been retrieved/built/added (CC @CDicksonG). This PR adds a spinner and also blocks sending whilst a grid link is being converted...
show-grid-link-conversion-spinner
(note GIF was recorded whilst network was severely throttled so you can see spinner for longer - it's normally much faster obvs.)

### DIFF
--- a/client/gql.ts
+++ b/client/gql.ts
@@ -261,10 +261,10 @@ export const gqlGetGridSearchSummary = gql`
     }
 `;
 
-export const gqlAsGridPayload = (gridUrl: string) => gql`
-    query AsGridPayload {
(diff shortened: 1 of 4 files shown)


twrichards:
> Nit: in the GIF above _please wait a moment…_ is wrong font, and a bit too big? Also, immediately after there is another _loading…_? Maybe make it exactly same-looking _loading… too?

addressed @paperboyo - GIF updated

PR #269, opened 2023-06-02
fix unsafe `reduce` on the result of `listLastItemSeenByUsers`

twrichards:
Had a few issues relating to this pop-up in Sentry so thought I'd fix


simonbyford (approved):
Looks good 👍

PR #271, opened 2023-06-06
instant emails for group mentions

twrichards:
Run `yarn database-setup`, selecting the stage and then the last option (to effectively run `019-TriggerEmailLambdaAfterItemInsert.sql`) for
- [x] **`CODE`**
- [x] **`PROD`**

---

In #259 we added the `email-lambda` but with only a special case for central production mentions, then in #268 we made emails for claims clearer and in #266 we turned on emails for missed individual mentions. All with the lambda being invoked by a schedule (every 5 mins), which makes sense for individual mentions as these go out if they've remained unread for an hour (currently) but for the central prod mentions and subsequent claims this can be a little slow...

SO now **all group mentions** are emailed almost instantly, by adding a DB 'after insert' trigger (like we have for desktop/push notifications) to send the item id of items with group mentions to the `email-lambda` as soon as they're created. For that item ID it receives, the `email-lambda` then performs a similar DB lookup as before to build up all the info it needs for the email send (rather than building it in the DB trigger, as that's trickier and harder to maintain), before sending out. 

The subject and email copy is slightly adjusted for the group mention emails, since they're instant (and 'missed' language didn't make sense)...
image
... note that we place the item id of the mention item in the subject as a ref, and then for any claims we place the `relatedItemId` in the subject as the ref (therefore the same subject as the original request email) so they end up threaded (i.e. claim email is a reply to the original)...
image

## How to test

With this branch deployed to `CODE` simply mention a group (that you're a member of) in any pinboard chat (making sure to click Yes on the modal asking whether it should be a request), see the email arrive promptly, then claim the item in pinboard and see a further email arrive as a **reply** stating that you've claimed the request.


twrichards:
Thank you @pradasa 

> I would probably change it to "your team has just been mentioned in Pinboard" or " your team has just been mentioned" (the emoji helps us not having to necessarily say "in Pinboard". It's a bit shorter and it will give us more space on the subject line before it gets truncated.

Will do (although I think I'll keep the `in Pinboard` bit, as more explicit than the emoji and I'm not too concerned if that's the bit to get truncated).

> Regarding the reference number, I wonder if that is clear to users. We can ask CP. My initial instinct would be to use the HL (if available) or WT as backup.

The ref is merely in the subject to control threading (to make it threaded/replies in the case of group mentions, and to prevent it in the case of individual missed mentions - where threading would cause confusion, since threading is based solely on subject line unfortunately). 
I did consider using the WT/HL in the subject, but decided against because they can change between the original mention email and the claim reply.

pradasa:
> The ref is merely in the subject to control threading (to make it threaded/replies in the case of group mentions, and to prevent it in the case of individual missed mentions - where threading would cause confusion, since threading is based solely on subject line unfortunately).
I did consider using the WT/HL in the subject, but decided against because they can change between the original mention email and the claim reply.

That makes sense. Thanks, Tom!

aracho1 (approved):
Tested in CODE and working beautifully 💯

PR #276, opened 2023-06-12
always capture `composerId` tag for `MESSAGE_SENT` telemetry events

twrichards:
Currently `composerId` is captured on all telemetry events where a `<pinboard-preselect ... />` element is present in the host application (i.e. when viewing a specific piece in composer).
For sends at least it would be better if we could attribute all send events (`MESSAGE_SENT`) with the `composerId` even if using Pinboard in grid, workflow etc. Since we have this information available already (with working title and headline retrieved from workflow via `workflow-bridge-lambda` periodically) in this PR we add `composerId` to the `tags` of the telemetry event in `SendMessageArea` component.

This will help us track progress towards our OKR; the ratio of pieces published to pieces where pinboard was used.

### Testing checklist
Send from...
- [x] specific composer piece (should work as before)
- [x] composer dashboard
- [x] grid
- [x] workflow (only after https://github.com/guardian/workflow-frontend/pull/403) 
- [x] 'peeking' at a team pinboard (so it's not part of your 'My Pinboards' at the point of send)

### DIFF
--- a/client/src/panel.tsx
+++ b/client/src/panel.tsx
@@ -288,6 +288,12 @@ export const Panel: React.FC<IsDropTargetProps> = ({ isDropTarget }) => {
           <Pinboard
             key={pinboardId}
             pinboardId={pinboardId}
+            composerId={useMemo(
+              () =>
+                activePinboards.find((_) => _.id === pinboardId)?.composerId ||
+                null,
+              [activePinboards, pinboardId]
+            )}
             isExpanded={pinboardId === selectedPinboardId && isExpanded}
             isSelected={pinboardId === selectedPinboardId}
             panelElement={panelRef.current}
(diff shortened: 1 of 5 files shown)


client/src/panel.tsx
```diff
@@ -288,6 +288,10 @@ export const Panel: React.FC<IsDropTargetProps> = ({ isDropTarget }) => {
           <Pinboard
             key={pinboardId}
             pinboardId={pinboardId}
+            composerId={
+              activePinboards.find((_) => _.id === pinboardId)?.composerId ||
```
jonathonherbert, line 292: Definitely non-blocking – could this be a bit slow with large numbers of pinboards, or is that unlikely? Conscious that this will run every render IIUC, and unsure how often that will happen given pinboard's a real time tool – could e.g. memoise tuple of active pinboard ids and composer ids if that was a concern.

jonathonherbert (approved):
Code looks good, looks like you've tested well in CODE so 👍  – one non-blocking comment.

client/src/panel.tsx
twrichards: good spot, memoised that now

PR #277, opened 2023-06-14
add message 'replies'

twrichards:
https://trello.com/c/KGTMhU1e/998-add-a-reply-feature-to-pinboard

Full-blown threading would be far too busy/confusing in this tightly constrained area of screen real estate, but for busy Pinboard conversations with multiple topics being discussed it's important to provide a way for users to be specific about what exactly they're responding to to avoid misunderstandings & mistakes. Hence a 'WhatsApp style reply' seemed like a good balance.

For some time we've had all the building blocks for 'replies' in Pinboard; the `NestedItemDisplay` added as part of 'claiming' functionality (see #205) and the 'hover menu' added for 'edit & delete' (see #230) - so most of this PR is just wiring the bits together 🎉 with a touch of refactoring.

Aside from some slight styling tweaks, the only main thing added on top of the existing functionality was payload thumbnails within the `NestedItemDisplay` (but aside from CSS scale, largely re-uses existing code/components - yay).

replies

### TODO
- [x] styling & copy tweaks most likely (CC @pradasa)
- [x] render mini thumbnail for any payloads in the `NestedItemDisplay` (useful if for example multiple images have been shared and someone wants to comment on a specific one, having a tiny thumbnail will help to avoid confusion


pradasa:
This looks great, @twrichards 
I'll have a look tomorrow AM. 

I think it makes sense to add it where you have it. One thing to consider is that at the moment you are replying to your own messages, which means that the edit and bin icon also appear. 
The most likely scenario is that a users would reply to someone else's scenario, so the reply icon would be the only one to appear. 

It does make sense to also have the option to reply/reference your own messages (as you have it), just like you can in Whatsapp

pradasa:
In terms of UI, could we please
- move the "in reply to" next to the timestamp. In the same style as we do "edited"
- change all text that is currently in italics to regular (including "edited", which I had missed)

We want to avoid using italics in general for UI best practices. Having the background rectangle behind the quoted message already tells us that the content is different to the message the user is writing so we don't need italics as well.

Thanks!

client/src/grid/gridDynamicSearchDisplay.tsx
```diff
@@ -164,10 +173,14 @@ export const GridDynamicSearchDisplay = ({
                   transition: 0s;
                 }
               `}
-              onClick={(event) => {
-                event.stopPropagation();
-                getGridSearchSummaryQuery.refetch();
-              }}
+              onClick={
+                shouldNotBeClickable
+                  ? undefined
```
aracho1, line 178: When should Grid assets not be clickable?

aracho1 (approved):
Tested in CODE, looks good to me 👍

PR #281, opened 2023-06-15
only 'eat' query params if they're actually present to avoid unnecessarily re-writing host application's URL

twrichards:
Quick tweak to #279 as I noticed pinboard was rewriting query params unnecessarily under normal operation of say Grid.


twrichards:
Tested in `CODE`

PR #282, opened 2023-06-21
tweaks to 'replies'

twrichards:
Two minor additions to https://github.com/guardian/pinboard/pull/277

- ensure changes to 'user lookup' are propagated to `NestedItemDisplay` to avoid emails being displayed on first load, e.g. 
image

- recognise claim items in NestedItemDisplay (because users might reply to the claim item), it previously displayed nothing for replies to claim items, e.g. 
image


frederickobrien (approved):
Very lovely

PR #283, opened 2023-06-26
add support for `mam-video` items (i.e. videos from `media-atom-maker`)

twrichards:
https://trello.com/c/SiW7WLsL/1513-pinboard-mam-videos-support

related: https://github.com/guardian/media-atom-maker/pull/1120

Following in footsteps of `grid-crop`, `grid-original` etc. in this PR we add support for `mam-video`. This involves 
- a new payload definition (which is similar to those for grid, but with an `externalUrl` property which holds an Youtube embed link)
- a new component `MamVideoDisplay` (with new video icon) which if the externalUrl is provided then renders that embed in a small iframe (just like MAM does itself) else falls back to displaying the thumbnail (the poster image in MAM)

... simple as that really.

mam-videos-in-pinboard

EDIT: since the GIF was made the `Add to 📌` button has been moved to be on the `Video Preview` heading line
image


aracho1 (approved):
LGTM

aracho1:
Non blocking as it still works fine and observed similar thing with Grid assets but the thumbnail overflows when the panel is re-sized too small, wonder if it's worth restricting how small the panel can be resized to 
image

PR #284, opened 2023-06-26
detect/convert MAM video links when pasted

twrichards:
https://trello.com/c/SiW7WLsL/1513-pinboard-mam-videos-support

Following on from #283 

Given there exists some muscle memory for sharing atoms using links, let's detect them and convert them to proper `mam-video` payloads. Here we detect MAM links and convert them to their corresponding MAM API urls and fetch (relies on CORS tweak in https://github.com/guardian/media-atom-maker/pull/1120) then we piece together the `externalUrl`, `thumbnail` and `embeddableUrl` (the same as via the `asset-handle` in https://github.com/guardian/media-atom-maker/pull/1120) replacing the pasted link if the payload can be built successfully.

Obviously this is similar to #257 but importantly calls MAM directly (AJAX) rather than via AppSync, since this didn't warrant a `mam-bridge-lambda` just for this purpose (but we had it already for grid, so that made sense for the link resolution).

mam-link-detection


aracho1 (approved):
✨

PR #286, opened 2023-07-03
some Firefox fixes

twrichards:
User reported scroll issues in Firefox which I was able to replicate, after some trial and error settled on a fix.
image

Also noticed and fixed some aliasing issues around the avatar roundel and some line-height issue.

Need to get better at testing on FF (not just Chrome where we're developing).


paperboyo (approved):
Testing positive! Thanks!

PR #289, opened 2023-07-20
throw error when real interaction happens against demo pinboard

twrichards:
Recently the `email-lambda` choked (and repeatedly failed every 5mins on the schedule) because it was trying to process an item where the `pinboardId` was `DEMO`, such items should never reach the database. I haven't yet been able to spot how that was possible in the logic, so I've added some safeguards for create, edit and delete of items which detects/throws when the tour isn't running but it still attempts a real mutation with a `pinboardId` of `DEMO` (so we'll have more info in Sentry) and it won't progress to perform the mutation.

According to the database it's happened 3 times in PROD to date (only the most recent had a mention and we noticed as a result of `email-lambda` errors).

### DIFF
--- a/client/src/editItem.tsx
+++ b/client/src/editItem.tsx
@@ -10,6 +10,7 @@ import { agateSans } from "../fontNormaliser";
 import { composer } from "../colours";
 import { PINBOARD_TELEMETRY_TYPE, TelemetryContext } from "./types/Telemetry";
 import { useTourProgress } from "./tour/tourState";
+import { demoPinboardData } from "./tour/tourConstants";
 
 interface EditItemProps {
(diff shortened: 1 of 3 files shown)


andrew-nowak (approved):
lgtm, though I wonder if this kind of validation would work better on the server somewhere?

twrichards:
> lgtm, though I wonder if this kind of validation would work better on the server somewhere?

yeah, we could fail in the `database-bridge-lambda` too/instead but Sentry would have the best information and we've already prevented it from causing the one problem we know about if it did make it through in #288 - so decided to stick with this approach for now. thanks for the review @andrew-nowak

PR #295, opened 2023-08-18
improve workflow integration

twrichards:
https://trello.com/c/Pw4EGPHn/1660-improve-pinboard-workflow-integration

Facilitated by https://github.com/guardian/workflow-frontend/pull/404

Previously pinboard integration in workflow was just an optional column displayed to permissioned users with an 'inline mode' for displaying the pinboard for the relevant row in workflow, introduced in #219 ). However we've had multiple bits of user feedback from users who spend lots of time in workflow, that they were missing unread messages from pinboards not displayed in workflow given their set of filters (e.g. an unrelated board in 'my pinboards').

This PR makes the floaty and panel display always in workflow (for permissioned users, facilitated by https://github.com/guardian/workflow-frontend/pull/404) but with some key tweaks to the SelectPinboard view, so that clicking a pinboard in `My Pinboards` or `My Teams' Pinboards` will...

- scroll to the relevant row in workflow and open the 'inline mode panel' for that pinboard IF the row is visible given current workflow filters
- ELSE open in a new tab to avoid confusion

... this solution represents a compromise of familiarity for users and simplicity of implementation whilst minimising the confusion that could arise from the regular Pinboard floaty/panel displaying the same items as displayed in workflow.

Given we now display the floaty all the time, there's now a warning about the column not being enabled with a handy button to enable the column (implemented by driving the underlying workflow interface 😈).

Also, this has meant we could re-instate the interactive demo in workflow (previously disabled because no floaty and so no SelectPinboard view so didn't make sense), which required a few tweaks (see relevant commit).

improved_workflow

### DIFF
--- a/client/fontNormaliser.ts
+++ b/client/fontNormaliser.ts
@@ -18,15 +18,17 @@ const agateFontNameVariants = [
   "GuardianAgateSans1Web",
 ];
 
+export const agateFontFamily = `${agateFontNameVariants
+  .map((_) => `"${_}"`)
+  .join(",")}, Arial, sans-serif;`;
+
 // source foundations will give us Guardian Text Sans, but we usually want Guardian Agate Sans, so as to fit in with
 // the tools hosting pinboard.
 const overrideToAgateSans: FontOverride =
   (originalFunc: FontScaleFunctionStr) => (options?: FontScaleArgs) =>
     `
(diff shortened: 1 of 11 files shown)


Georges-GNM (approved):
Think this is good! Ran it with the workflow PR as well and seemed to work as intended 👍

PR #296, opened 2023-08-29
Rewrite React.fc<type> props declarations

Georges-GNM:
## What does this change?

As suggested in https://github.com/guardian/pinboard/pull/295#discussion_r1309051825, this rewrites various props declarations from:
`export const AssetView: React.FC<AssetView> = ({ items }) => {...}`
to: 
`export const AssetView = ({ items }: AssetViewProps) => {...}`

Functionally this is the same, but this seems like a nicer way to write it. 

## How to test

Since this is a small code style refactor, the project should still compile and intellisense in vscode or intellij should keep picking up on types for the specified props:

image

### DIFF
--- a/client/src/globalState.tsx
+++ b/client/src/globalState.tsx
@@ -4,7 +4,13 @@ import {
   useMutation,
   useQuery,
 } from "@apollo/client";
-import React, { useCallback, useContext, useEffect, useState } from "react";
+import React, {
+  ReactNode,
+  useCallback,
+  useContext,
+  useEffect,
+  useState,
+} from "react";
 import { Item, MyUser } from "../../shared/graphql/graphql";
 import {
   gqlAddManuallyOpenedPinboardIds,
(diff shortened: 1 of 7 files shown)


client/src/globalState.tsx
```diff
@@ -144,7 +152,7 @@ export const GlobalStateProvider: React.FC<GlobalStateProviderProps> = ({
   hasEverUsedTour,
   visitTourStep,
   children,
-}) => {
+}: GlobalStateProviderProps) => {
```
twrichards, line 155: I think if you do something like the following you can avoid the need for an explicit `children: ReactNode;` in the `GlobalStateProviderProps` interface...
```suggestion
}: PropsWithChildren<GlobalStateProviderProps>) => {
```

Alternatively, rather than `interface GlobalStateProviderProps {` further up, I think you could write `type GlobalStateProviderProps = PropsWithChildren<{`

twrichards (approved):
great thank you for this @Georges-GNM 👏 

one [non-blocking] suggestion

client/src/globalState.tsx
Georges-GNM: Ah yeah, that seems to work! Good shout, thanks

PR #298, opened 2023-10-20
support soft-breaks in messages 

twrichards:
Legit request from user...

> How about allowing soft returns in the typing box so you can separate different queries or put in the relevant quote without running on? A soft return currently means "send".

I will push back on the first example usage of soft breaks and encourage separate queries to intentionally be different messages, so they can be 'replied' to (see https://github.com/guardian/pinboard/pull/277). 

So this PR...

- detects shift/alt when pressing Enter key and doesn't send accordingly (allowing the default behaviour of soft line break)
- respect line breaks when rendering (with `white-space: pre-line`)

soft

### DIFF
--- a/client/src/itemDisplay.tsx
+++ b/client/src/itemDisplay.tsx
@@ -251,7 +251,13 @@ export const ItemDisplay = ({
                       userLookup,
                     ]
                   )}
-                {formattedMessage}
+                <span
+                  css={css`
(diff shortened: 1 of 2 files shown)


Georges-GNM (approved):
LGTM, tested in locally and could confirm the behaviour 👍

PR #299, opened 2023-10-20
trim ✂️ 

twrichards:
Little follow-up on #298 

- prevent sending message if composed entirely of whitespace
- trim surrounding whitespace when sending messages

trim


Georges-GNM (approved):
✂️

PR #301, opened 2024-01-24
Add devx-backup tag to database

Georges-GNM:
## What does this change?

Part of [RDS Backups](https://trello.com/c/ZQ2cQWQD/2032-rds-backups-devx-pairing) 

Enables more robust backup processes powered by AWS backup service for the RDS database, as per this [FAQ](https://docs.google.com/document/d/1VDCSxYFlWs4R6g0Waa6OmmfytV60AROyHxfIGho7cLA/edit#heading=h.vwt7syo8ng40).

## How to test

This was added to code, and we could verify this successfully enabled the desired backup in the AWS console.

### DIFF
--- a/cdk/lib/stack.ts
+++ b/cdk/lib/stack.ts
@@ -18,6 +18,7 @@ import {
   Fn,
   RemovalPolicy,
   Stack,
+  Tags,
 } from "aws-cdk-lib";
 import * as appsync from "@aws-cdk/aws-appsync-alpha";
 import { join } from "path";
@@ -110,6 +111,7 @@ export class PinBoardStack extends GuStack {
       publiclyAccessible: false,
       removalPolicy: RemovalPolicy.RETAIN,
     });
+    Tags.of(database).add("devx-backup-enabled", "true");
 
     const roleToInvokeLambdaFromRDS = new iam.Role(
       this,
--- a/cdk/lib/__snapshots__/stack.test.ts.snap
+++ b/cdk/lib/__snapshots__/stack.test.ts.snap
@@ -617,6 +617,10 @@ Object {
         "StorageType": "gp2",
         "Tags": Array [
           Object {
+            "Key": "devx-backup-enabled",
+            "Value": "true",
+          },
+          Object {
             "Key": "gu:cdk:version",
             "Value": "49.5.0",
           },
@@ -721,6 +725,10 @@ Object {
         },
         "Tags": Array [
           Object {
+            "Key": "devx-backup-enabled",
+            "Value": "true",
+          },
+          Object {
             "Key": "gu:cdk:version",
             "Value": "49.5.0",
           },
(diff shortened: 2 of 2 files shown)


michaelwmcnamara (approved):

cdk/lib/stack.ts
```diff
@@ -110,6 +111,7 @@ export class PinBoardStack extends GuStack {
       publiclyAccessible: false,
       removalPolicy: RemovalPolicy.RETAIN,
     });
+    Tags.of(database).add("devx-backup-enabled", "true");
```
michaelwmcnamara, line 114: Looks like this db instance instantiates a number of different objects and the 'devx-backup-enabled' tag is added to all of these rather than just the db engine. Hold off on merging this until we verify that this won't have any negative impacts.

michaelwmcnamara (changes requested):
Holding off on merge until we have verified according to above comment.

jacobwinch (approved):
Although it'd be cleaner to back up the `AWS::RDS::DBInstance` resource type only, I don't think there is any real downside to adding this tag to the associated resources, which are of the following types:

* `AWS::IAM::Role`
* `AWS::EC2::SecurityGroup`
* `AWS::RDS::DBSubnetGroup`
* `AWS::SecretsManager::Secret`

IIUC AWS Backup [doesn't understand any of these types anyway](https://docs.aws.amazon.com/aws-backup/latest/devguide/working-with-supported-services.html), so the backup tooling is not going to select them.

In other words, I think it's fine to merge this as is.

Update: if we really want to access the RDS instance only for tagging purposes, we have to access the [L1 construct](https://docs.aws.amazon.com/cdk/v2/guide/constructs.html#constructs_l1_using) - there's an example of that here: https://github.com/guardian/pinboard/pull/302.

Georges-GNM:
> Although it'd be cleaner to back up the `AWS::RDS::DBInstance` resource type only, I don't think there is any real downside to adding this tag to the associated resources, which are of the following types:
> 
> * `AWS::IAM::Role`
> * `AWS::EC2::SecurityGroup`
> * `AWS::RDS::DBSubnetGroup`
> * `AWS::SecretsManager::Secret`
> 
> IIUC AWS Backup [doesn't understand any of these types anyway](https://docs.aws.amazon.com/aws-backup/latest/devguide/working-with-supported-services.html), so the backup tooling is not going to select them.
> 
> In other words, I think it's fine to merge this as is.
> 
> Update: if we really want to access the RDS instance only for tagging purposes, we have to access the [L1 construct](https://docs.aws.amazon.com/cdk/v2/guide/constructs.html#constructs_l1_using) - there's an example of that here: #302.

As we've tested this particular implementation on code and were satisfied it works (and would also like to get this merged now to verify prod tomorrow), I'll go ahead and merge this one. If there's any issues caused by the tagging on the other resources, it's good to know how we can get it to be limited to the instance though.

PR #303, opened 2024-01-26
Update RDS database cdk

Georges-GNM:
## What does this change?

Makes the config slightly safer by adding deleteProtection and enabling us to delete backups in a straightforward manner if needed. 

## How to test

This has been put on code, we could verify the cfn was appropriately updated.

### DIFF
--- a/cdk/lib/__snapshots__/stack.test.ts.snap
+++ b/cdk/lib/__snapshots__/stack.test.ts.snap
@@ -593,6 +593,7 @@ Object {
         "DBSubnetGroupName": Object {
           "Ref": "DatabaseSubnetGroup7D60F180",
         },
+        "DeleteAutomatedBackups": false,
         "DeletionProtection": true,
         "EnableIAMDatabaseAuthentication": true,
         "Engine": "postgres",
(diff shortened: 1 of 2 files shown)


cdk/lib/__snapshots__/stack.test.ts.snap
```diff
@@ -593,6 +593,7 @@ Object {
         "DBSubnetGroupName": Object {
           "Ref": "DatabaseSubnetGroup7D60F180",
         },
+        "DeleteAutomatedBackups": false,
         "DeletionProtection": true,
```
jacobwinch, line 597: Interesting that `DeletionProtection` is already `true` in the CFN diff - looks like it is [already being set](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_rds.DatabaseInstance.html#deletionprotection) due to:

```typescript
removalPolicy: RemovalPolicy.RETAIN
```

twrichards: Interesting! I'd like to keep the 'deletionProtection' explicit though (i.e. this PR as-is)

twrichards (approved):

PR #307, opened 2024-05-15
 Update yarn to latest stable version

mchv:
## What does this change?

This update `yarn` from `1.22.17 `to the latest stable version `4.2.2`.

There are a few noteworthy changes:

#### Adding the range for `react-dom` and `react` dependencies 

The parsing of version dependency since yarn `2.x` [is more strict](https://github.com/yarnpkg/berry/issues/1621), which  means the `react` and `react-dom` `alias package` needs an additional range to work ([preact issue about it](https://github.com/preactjs/preact-www/issues/1050))

#### Removal of `yarn-duplicate`

The [yarn-duplicate package](https://www.npmjs.com/package/yarn-deduplicate) is not necessary anymore, and has been replaced by the [native dedupe command](https://yarnpkg.com/cli/dedupe):
 - commit f7abd45833682923b15e00c3b1bba4314b7a9b91
 - commit b8b0617254f16e1e60c41d2195fe9435ce3f6c06
 - commit 71195d9fe9872bdf12d8b5334b28f9dfc808ef57

#### Warnings displayed when installing

When executing `yarn install` a few warnings are now displayed  

Screenshot 2024-05-15 at 21 13 22

The associated changes do not address them, but with the associated changes it should be easier to update packages to address them (for instance updating `jest` and `ts-jest` seems sensible and not too difficult)

#### Retrieval of the right version of `yarn`

In order to retrieve the right version of yarn, it is now recommended to use [corepack](https://yarnpkg.com/corepack), rather than  
This is very similar to [that change in dotcom-rendering](https://github.com/guardian/dotcom-rendering/pull/9610):
 - commit c63df94a2358e955668d57a0574095f233a9474f 

#### Upgrade of node version 

However for the right `yarn` version to be picked up, un upgrade to the latest `18.xx` node version was needed (see https://github.com/pnpm/pnpm/issues/7724), otherwise yarn version is not found:

And `yarn install` fail miserably:
```
Type Error: URL.canParse is not a function
    at parseSpec (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:23627:21)
    at loadSpec (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:23704:11)
    at async findProjectSpec (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:23649:20)
    at async executePackageManagerRequest (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:24223:18)
[...]

PR #314, opened 2024-10-22
fix husky `pre-commit` hook

twrichards:
run `fnm use` and `nvm use` to ensure correct version of node before executing the various steps in the `pre-commit` hook (configured via husky)

Broken since #307

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,7 +1,14 @@
 #!/bin/sh
+echo "Running pre-commit hook"
+
 . "$(dirname "$0")/_/husky.sh"
 
 set -e
+
+fnm use
+
+fnm use || true
+nvm use || true
 
 corepack enable
 yarn lint-staged


.husky/pre-commit
```diff
@@ -1,8 +1,15 @@
 #!/bin/sh
+echo "Running pre-commit hook"
+
 . "$(dirname "$0")/_/husky.sh"
 
 set -e
```
bryophyta, line 7: why do we have `fnm use` twice?

twrichards: wooops, that's a rebase issue, fixing now, thanks for the spot @bryophyta

bryophyta:
the code looks good to me but I haven't been able to trigger the hook itself...

```
./scripts/setup.sh
touch temp_file
git add temp_file
git commit -m "Test commit"
git push
```

Am I missing a step to get the hooks set up?

twrichards:
> the code looks good to me but I haven't been able to trigger the hook itself...
> 
> ```
> ./scripts/setup.sh
> touch temp_file
> git add temp_file
> git commit -m "Test commit"
> git push
> ```
> 
> Am I missing a step to get the hooks set up?

its a pre-commit hook, so shouldn't need the git push (basically the commit should be slow), not sure why it's not running for you though, will message you offline...

twrichards:
@bryophyta I can't seem to think why the hook wasn't running for you. I've just updated the `setup.sh` script which might help a little

If that doesn't work, perhaps you might try https://typicode.github.io/husky/troubleshoot.html

andrew-nowak (approved):
lgtm

PR #315, opened 2024-10-24
Bump Node.JS to v20.18.0

davidfurey:
Minimal changes to upgrade to NodeJS 20. Node.JS 18 has an AWS Lambda deprecation date of Jul 31, 2025. Getting ahead of this since we've done a lot of Node.JS upgrades recently.

Requires aws-cdk version bump because 2.68.0 doesn't have `lambda.Runtime.NODEJS_20_X`


twrichards (approved):
I deployed to `CODE` and gave it a spin, everything seems to be working.

PR #317, opened 2024-10-29
replace deprecated `aws-serverless-express` with `@codegenie/serverless-express`

twrichards:
Since #315 we saw increased latency and 5XX from the bootstrapping-lambda (& its API Gateway) such that pinboard loaded flakily/intermittently for users. Logs weren't particularly revealing, but hinted at something related to [`aws-serverless-express`](https://www.npmjs.com/package/aws-serverless-express), which is deprecated. So, in this PR replace with [`@codegenie/serverless-express`](https://www.npmjs.com/package/@codegenie/serverless-express), which we've has been working nicely in https://github.com/guardian/octopus (serving a similar purpose).

The 5XX issues have been difficult to replicate in CODE, so we may find that we need to test this in PROD and keep a close eye on metrics for a few hours after release.

Note, alarm to catch 5XXs is added in https://github.com/guardian/pinboard/pull/316

### DIFF
--- a/bootstrapping-lambda/package.json
+++ b/bootstrapping-lambda/package.json
@@ -16,7 +16,6 @@
     "@aws-sdk/client-lambda": "^3.299.0",
     "@aws-sdk/client-s3": "^3.299.0",
     "@babel/preset-typescript": "^7.18.6",
-    "@types/aws-serverless-express": "^3.3.3",
     "@types/cors": "^2.8.12",
     "@types/express": "4.17.17",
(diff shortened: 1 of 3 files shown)


andrew-nowak (approved):

twrichards:
this looks to have done the trick in PROD, no noticeable increase in 5XX.

it does appear to add approx 150ms to `bootstrapping-lambda` duration (and therefore latency of requests to load pinboard, via the API Gateway) but this is tolerable

PR #318, opened 2024-10-30
increase bootstrapping-lambda from 128MB to 256MB

twrichards:
The node20 upgrade in #315 introduced some instability in bootstrapping-lambda, and whilst #317 was a worthy upgrade of a deprecated package it didn't fix the problems as hoped. 

In CODE after hammering `/pinboard.loader.js` I was able to replicate the issues, and then after some reading, experimented with increasing memory (as that also proportionally increases CPU) and which made a decent reduction on duration and haven't yet been able to replicate the issues, so in this PR I codify the change so we can try out in PROD (it still might not fix things if it's a memory leak it might just take longer to manifest but worth a try nonetheless).

### DIFF
--- a/cdk/lib/__snapshots__/stack.test.ts.snap
+++ b/cdk/lib/__snapshots__/stack.test.ts.snap
@@ -3554,7 +3554,7 @@ $util.toJson($ctx.result)",
         },
         "FunctionName": "pinboard-bootstrapping-lambda-TEST",
         "Handler": "index.handler",
-        "MemorySize": 128,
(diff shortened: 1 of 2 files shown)


twrichards:
Curiously when hammering in CODE (with this branch deployed) this is quite typical 
image
... so I think we were just on the limit of memory. This is further confirmed by looking at the log lines from PROD prior to the Node20 upgrade (#315) ... 
image

davidfurey (approved):
Seems sensible to me

PR #321, opened 2024-11-05
add feature flag mechanism

twrichards:
As a pre-requisite to #312 we needed a way for users to gradually enrol into the new functionality as they receive 'training', which needed to persist across machines and work consistently in different tools, so storing a new column in the DB seemed sensible ...

- new JSONB column, called `featureFlags` on the `User` table
- becomes a field of the `MyUser` type (just a JSON string, parsed in the client)
- parsed `feautureFlags` shared around the app via the GlobalStateContext
- new GraphQL mutation `changeFeatureFlag` which writes key/values to the featureFlags JSONB column
- consumes changes to feature flags via query params with prefix `pinboardFeatureFlag_`, dropping them from URL once the new mutation `changeFeatureFlag` has been called with the new value (`true` or `false`) -- this allows trainers to provide a URL with feature flag query param to users and then it will remain on for the user indefinitely following their training
- renames/broadens the GraphQL subscription `onManuallyOpenedPinboardIdsChanged` to become `onMyUserChanges` and now fires based on the new mutation too, which allows all connected clients to hear about changes to featureFlags and react accordingly

### DIFF
--- a/cdk/lib/__snapshots__/stack.test.ts.snap
+++ b/cdk/lib/__snapshots__/stack.test.ts.snap
@@ -2475,6 +2475,7 @@ type Mutation {
   addManuallyOpenedPinboardIds(pinboardId: String!, maybeEmailOverride: String): MyUser
   removeManuallyOpenedPinboardIds(pinboardIdToClose: String!): MyUser
   visitTourStep(tourStepId: String!): MyUser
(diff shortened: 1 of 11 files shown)


andrew-nowak (approved):
code lgtm
tested on code, also looks good, though worth noting that I initially get "undefined" as the value of my feature flag
image
I think that's expected behaviour, but may need bearing in mind for any usages of the feature flags

PR #325, opened 2024-11-22
consolidate subscriptions per active pinboard into a few subscriptions at the top level AND eliminate worst/expensive forms of polling

twrichards:
## Background 

In response to a user being effectively blocked from using pinboard due to reaching 100 AppSync subscriptions (a limit we were previously unaware of), we raised #323 to cap the number of 'my pinboards' _mounted_ to 50 as each one had a subscription to listen to messages for that pinboard, but this did not solve the problem as it turns out each pinboard actually had 3 subscriptions (one for items, one for claims and one for last item seen) and the user had 34 'my pinboards' 🤦  (the change was therefore reverted in #324 as it was ineffective).

Whilst a user having so many 'my pinboards' is really a symptom of poor housekeeping in workflow... it did make me re-evaluate the idea of 3 subscriptions per mounted pinboard... which I concluded was pretty wasteful.

## Consolidating Subscriptions
So now instead of 3 subscriptions PER 'my pinboard' we now have 3 _shared_ subscriptions up in the global state, listening to ALL new items, ALL claims and ALL last item seen, then storing to state only those which are relevant to any of the 'my pinboards' and then individual instances of `Pinboard` pull from the global state and filter for the ones with matching `pinboardId` - this is much more efficient.

## Eliminating some polling
With the shared subscriptions in place, listening for everything, we can now keep a count (running total) of all items created/changed/claimed in ANY pinboard, and use that to trigger the `getGroupPinboardIds` and `getItemCounts` queries previously performed on a polling frequency (see #304 for when polling frequency was reduced to save money). 

Since the polling is 15s currently (see #304) is now it's being replaced with it being triggered by things actually being sent/claimed **we should expect things to be more realtime** (by up to 15s improvement effectively).

### Potential/Probable Cost Savings
We know that AppSync queries make up the vast majority of the cost of running pinboard...

image

... we know we **average around 1.7millions AppSync requests per day** currently (down from 3.7mill per day prior to #304), and of those requests `getGroupPinboardIds` and `getItemCounts` make up the vast majority...

image

... but we also know that there are vastly fewer messages sent/claimed per day (around 50 per day) than the total number of AppSync requests...

image

... so we should see far fewer requests in total once this is release and therefore I am relatively confident that replacing these two expensive forms of polling we should see some significant cost savings too 💸 🎉 .

### DIFF
--- a/client/src/globalState.tsx
+++ b/client/src/globalState.tsx
@@ -566,7 +566,8 @@ export const GlobalStateProvider = ({
     }
     setSelectedPinboardId(null);
     setError(pinboardIdToClose, undefined);
-    //TODO probably need to check if pinboard is also in group pinboards, so we don't lose an unread on there
+
+    // note that panel.tsx detects this and reinstates any hasUnread from the group pinboards
     setUnreadFlag(pinboardIdToClose)(undefined);
   };
 
--- a/client/src/globalState.tsx
+++ b/client/src/globalState.tsx
@@ -67,6 +67,7 @@ interface GlobalStateContextShape {
   allSubscriptionItems: Item[];
   allSubscriptionClaimedItems: Item[]; // both the updated 'claimed' item and the new 'claim' item
   allSubscriptionOnSeenItems: LastItemSeenByUser[];
+  totalItemsReceivedViaSubscription: number;
 
   payloadToBeSent: PayloadAndType | null;
   setPayloadToBeSent: (newPayloadToBeSent: PayloadAndType | null) => void;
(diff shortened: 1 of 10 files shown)


twrichards:
@paperboyo and I just tested this, chatting to one another (me in workflow [where there WAS `getItemCounts` polling] and @paperboyo in composer [where there WAS `getGroupPinboardIds` polling) including making 'requests' via group mentions and claiming - everything worked nicely including unread counts etc all updated in near real-time 🎉  (IMO good to go for PROD release on Monday - pending review of course)

client/src/globalState.tsx
```diff
@@ -460,8 +565,9 @@ export const GlobalStateProvider = ({
       );
     }
     setSelectedPinboardId(null);
-    setUnreadFlag(pinboardIdToClose)(undefined);
     setError(pinboardIdToClose, undefined);
+    //TODO probably need to check if pinboard is also in group pinboards, so we don't lose an unread on there
```
bryophyta, line 569: is this a todo that should be addressed in this PR, or is it one for an optional followup?

twrichards: pinboard_group_unread_after_close

PR #327, opened 2024-11-29
add `Stage` to the namespace of the metric filter designed to catch the `email-lambda` race condition

twrichards:
Tiny tweak following #326 as I noticed CODE and PROD published to the same metric, as such if the issue occurred in either CODE or PROD, then both CODE and PROD alarms would fire - which is not what we want. This PR should address that.

### DIFF
--- a/cdk/lib/__snapshots__/stack.test.ts.snap
+++ b/cdk/lib/__snapshots__/stack.test.ts.snap
@@ -1632,6 +1632,12 @@ Object {
         },
         "MetricTransformations": Array [
           Object {
+            "Dimensions": Array [
+              Object {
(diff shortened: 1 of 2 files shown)


twrichards:
> failed to deploy - can't use dimensions with metric filters apparently - will just have to use stage in the metric name 😞

made stage a part of the metric namespace, with a `/` so it gets arranged in a folder (like AWS' metrics)...
image
(this doesn't show the folder layout as there needs be at least one event for it to show, and fortunately thanks to #326 we've not seen this issue again)

Fweddi (approved):

andrew-nowak (approved):

PR #328, opened 2024-12-05
upgrade to Node 22 (from 20)

twrichards:
Unfortunately #315 was just a touch before Node 22 support was announced (https://aws.amazon.com/blogs/compute/node-js-22-runtime-now-available-in-aws-lambda/)

So this PR moves from 20 to 22 - both for building and for the lambda runtimes.

✅  seems fine in `CODE`


bryophyta (approved):
lgtm and works locally 👍

PR #331, opened 2024-12-10
upgrade pan-domain-node to 0.5.0 to fix cookie expiry bug

andrew-nowak:
## What does this change?

Upgrades pan-domain-node to receive bugfix for cookie expiry date validation

See https://github.com/guardian/pan-domain-authentication/pull/167

### DIFF
--- a/bootstrapping-lambda/package.json
+++ b/bootstrapping-lambda/package.json
@@ -26,7 +26,7 @@
   },
   "dependencies": {
     "@codegenie/serverless-express": "^4.16.0",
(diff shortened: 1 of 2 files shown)


twrichards (approved):
the panda cookie is only checked when pinboard has loaded, which is after the host tool has loaded (in doing so already ensuring not expired) - so I can't this new/correct behaviour causing problems (in other words, I'm confident we were not relying on the faulty behaviour)

I've not had a chance to test in CODE yet, but if things load as normal - I think this is safe to merge.

PR #340, opened 2025-02-05
Remove deprecated Snyk sbt workflow

rhystmills:
## What does this change?

The Snyk `sbt` workflow is deprecated, and we have removed it [in other repos](https://github.com/guardian/grid/pull/4394).

I also don't think Pinboard has any Scala in it, or `sbt` dependencies (it's all TypeScript), so I'm unsure what value the workflow would provide if it weren't deprecated.

As an organisation are moving from Snyk to Dependabot for vulnerability management.

(Please let me know if I'm missing something!)


emdash-ie (approved):
Looks good! Agreed that we don't need to replace this as the repository doesn't have any scala or kotlin code (the only two cases DevX have mentioned where github's inbuilt dependency tracking doesn't work for us).

PR #342, opened 2025-02-20
add 1:1 (Square) as a suggestible ratio for composer trail images

twrichards:
Small tweak following on from #312 , to add `1:1` as a suggestible crop ratio (to support highlights containers for example), which makes it look like so (in composer)...
image


paperboyo (approved):
One-liners. Like. Match my reviewing abilities!

PR #345, opened 2025-03-12
explicitly send users to the push-notifications index.html page, as trailing slashes cause problems

andrew-nowak:
## What does this change?

When switching to the updated version of serverless-express, they have included breaking changes around how the path is built. It now uses the captured/proxy part of the route, which will not persist a trailing slash. This means that a request for `/push-notifications/` will be received by express.static as `/push-notifications`, which will return a redirect to `/push-notifications/`. But that's where the browser already is, so that'll just cause the same redirection to be issued, again, and again, and again...

This is a known issue in the new serverless-express: https://github.com/CodeGenieApp/serverless-express/issues/391, which has been attempted to be fixed previously: https://github.com/CodeGenieApp/serverless-express/pull/441, but of course a fix for a breaking change is itself a breaking change, and if it comes too slowly then users will have already built around the different behaviour, and clamour for a revert to fix their apps: https://github.com/CodeGenieApp/serverless-express/pull/455. Sigh.

This seems almost impossible to fix now, so opt out of it entirely by not deliberately sending a request for a trailing slash: return to the early web and make visible requests for an `index.html`.

## How to test

Deploy to CODE and attempt to toggle your notification settings.

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

### DIFF
--- a/client/src/pushNotificationPreferences.tsx
+++ b/client/src/pushNotificationPreferences.tsx
@@ -15,7 +15,7 @@ const toolsDomain = hostname.includes(".local.")
   ? "code.dev-gutools.co.uk"
   : "gutools.co.uk";
 
-export const desktopNotificationsPreferencesUrl = `https://pinboard.${toolsDomain}/push-notifications/`;
+export const desktopNotificationsPreferencesUrl = `https://pinboard.${toolsDomain}/push-notifications/index.html`;
 
 export const PushNotificationPreferencesOpener = ({
   hasWebPushSubscription,


SHession (approved):
LGTM! 

– Haven't run the code, but tried navigating to the path directly and it worked!

PR #349, opened 2025-03-18
trigger email-lambda immediately for individual mentions too (and simplify/refactor out the missed mentions logic)

twrichards:
Based on feedback as people use pinboard more (thanks to #312) people would like to be emailed immediately for individual mentions (just like teams are emailed immediately for group mentions). 

Overall this is a major simplification as it means we can eliminate the scheduling and logic to find/aggregate individual mentions which have gone unread for an hour.

**_I recommend reviewing with whitespace changes hidden_** 
image

#### CODE
- [x] re-run `019-TriggerEmailLambdaAfterItemInsert.sql` to update the trigger (to also include individual mentions)
- [ ] AFTER the `email-lambda` is deployed, then run `021-DropIsEmailEvaluatedColumnFromItemTable.sql`
#### PROD
- [ ] re-run `019-TriggerEmailLambdaAfterItemInsert.sql` to update the trigger (to also include individual mentions)
- [ ] AFTER the `email-lambda` is deployed, then run `021-DropIsEmailEvaluatedColumnFromItemTable.sql`

### DIFF
--- a/cdk/lib/__snapshots__/stack.test.ts.snap
+++ b/cdk/lib/__snapshots__/stack.test.ts.snap
@@ -24,9 +24,8 @@ Object {
       "GuCname",
       "GuCname",
       "GuCname",
-      "GuScheduledLambda",
+      "GuLambdaFunction",
       "GuLambdaErrorPercentageAlarm",
-      "GuAlarm",
       "GuAlarm",
     ],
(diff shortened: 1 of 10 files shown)


andrew-nowak (approved):
lgtm

twrichards:
this has been tested in PROD and is working nicely 🎉

PR #350, opened 2025-03-18
add more tolerance to the `pinboard-bootstrapping-lambda-api-CODE 5XX errors` alarm given CODE invocations are fewer and more sporadic

twrichards:
thanks to a nudge from @emdash-ie I was reminded that the 5XX alarm for CODE bootstrapping-lambda (introduced in #316 ) occasionally fires and we don't want people to ignore it - so this change now only fires the alarm if errors for longer (30mins CODE vs 10mins for PROD now)


emdash-ie (approved):
Much appreciated, thanks!

PR #354, opened 2025-03-20
use `width` rather than `x` to determine left vs right of the hover on `SuggestAlternateCrops`

twrichards:
…given the slide out animation on composer can make the x non-deterministic

A fix to https://github.com/guardian/pinboard/pull/353


twrichards:
Tested on CODE (more so than last time where I missed this jank in CODE)

andrew-nowak (approved):

PR #359, opened 2025-05-06
Remove option to suggest 5:4 alternatives

andrew-nowak:
## What does this change?

Now that 5:4 crops are our default, remove the ability to suggest them as "alternative" crops (as they're not alternative, they're the default!)


twrichards:
Might want to check with Jo Ruck etc. as I seem to recall asking if we'd want to get rid when I was prepping PRs and someone (Jo I think) said it would still be useful to be able to suggest alternates via this mechanism

PR #360, opened 2025-05-09
move 5:4 crop suggestion to end of list

andrew-nowak:
## What does this change?

Rather than remove this ratio from the list of suggestibles in #359 , we've been asked to keep, but send it to the bottom of the list. 

(Today I learned the properties in a javascript object are in fact ordered in the ES spec, by creation time unless keys are numeric -- unlike JSON where they are _not_ ordered. So simply changing the order in the code here is enough to get this to work)


bryophyta (approved):
lgtm 👍

PR #361, opened 2025-05-09
Allow "admins" to delete any pinboard message.

andrew-nowak:
## What does this change?

Occasionally a message or suggested image in pinboard should be disregarded, but the sender may not be available to remove it. Create a new permission that will grant CP and similar the ability to remove items that others have sent. We expect the permission to not be widely granted.

## How to test

Deploy to CODE along with https://github.com/guardian/permissions/pull/267. Send a message on pinboard, then get a friend, grant them the new permission, wait a minute or two then ask them to delete your message.


shared/permissions.ts
```diff
@@ -14,23 +14,84 @@ interface Permission {
...
+};
+
+export const userHasPermission = async (
+  userEmail: string,
+  permission: string
+): Promise<boolean> => {
+  const overrides = await getAllPinboardOverrides(permission);
+
+  return Boolean(
+    overrides?.find(({ userId, active }) => userId === userEmail && active)
```
bryophyta, line 95: nit: `some` feels slightly more semantic than `find` here?
```suggestion
    overrides?.some(({ userId, active }) => userId === userEmail && active)
```

bryophyta (approved):
Code changes look good to me, and functionality works as expected when running locally 👍

PR #362, opened 2025-05-12
Bumps the jump box AMI to Ubuntu 22

SHession:
## What does this change?

As Ubuntu 20 is no longer supported this PR bumps the jump box AMI to Jammy.  

## How to test

I'm not sure how to test this, I can find any documentation in the repo for making use of it. 

No EC2 instances or ASG referencing Ubuntu 20

This PR bumps the Java version from 8 -> 11. I don't expect this to be an issue.

### DIFF
--- a/riff-raff.yaml
+++ b/riff-raff.yaml
@@ -28,7 +28,7 @@ deployments:
         CODE: PinBoardStack-CODE.template.json
         PROD: PinBoardStack-PROD.template.json
       amiTags:
-        Recipe: editorial-tools-focal-java8-ARM-WITH-cdk-base
+        Recipe: editorial-tools-jammy-java11-ARM-WITH-cdk-base
         AmigoStage: PROD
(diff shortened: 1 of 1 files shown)


riff-raff.yaml
```diff
@@ -28,7 +28,7 @@ deployments:
         CODE: PinBoardStack-CODE.template.json
         PROD: PinBoardStack-PROD.template.json
       amiTags:
-        Recipe: editorial-tools-focal-java8-ARM-WITH-cdk-base
+        Recipe: editorial-tools-jammy-java11-ARM-WITH-cdk-base
```
andrew-nowak, line 31: suggest using `editorial-tools-jammy-java11` instead - from a look at the recipe they should be essentially equivalent, but with a more rationalised name now that ARM and cdk-base are standard in our recipes

andrew-nowak (approved):

PR #373, opened 2025-11-20
Fix glob vulnerability

liamness:
## What does this change?

Bumps `glob` to `10.5.0`. See:

- https://github.com/guardian/pinboard/security/dependabot/169
- https://github.com/isaacs/node-glob/issues/636

This does only look like a devDependency, so I'm not sure why we're seeing this on the grafana dashboard. But hopefully this will make the alert go away.

## How to test

Planning to deploy to CODE and then look at the lambda runs that result. Maybe this link will be useful (need to be logged into the workflow account): https://753338109777-vvxobf3n.eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logsV2:log-groups/log-group/%2Faws%2Flambda%2Fpinboard-archiver-lambda-CODE

Squashed alert

The regular archival job not working correctly? So test on CODE first.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -10023,17 +10023,18 @@ __metadata:
   linkType: hard
 
 "glob@npm:^10.2.2, glob@npm:^10.3.10":
-  version: 10.3.15
-  resolution: "glob@npm:10.3.15"
+  version: 10.5.0
+  resolution: "glob@npm:10.5.0"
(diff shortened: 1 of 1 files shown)


emdash-ie (approved):
Change looks good! Approving pending your testing in CODE

PR #374, opened 2025-12-02
Bumps GuCDK to 62.1.1

LAKSHMIRPILLAI:
## What does this change?

This PR bumps various CDK dependencies and removes the project's use of AWS SDK V2, covered in https://github.com/guardian/workflow/issues/1273.

## How to test

1. Deploy this branch to CODE 
2. Open Composer CODE
3. Open Pinboard and check if everything is working as usual and we are getting 200 response 

No usage of AWS SDK V2 and latest Guardian CDK changes.

Note: At one step , running  yarn dedupe --check   caused some failures saying -' 74 packages can be deduped using the highest strategy. Resolved this by running 'yarn dedupe --strategy highest' which updated the yarn.lock file

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -85,17 +85,10 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@aws-cdk/asset-awscli-v1@npm:^2.2.208":
-  version: 2.2.213
-  resolution: "@aws-cdk/asset-awscli-v1@npm:2.2.213"
-  checksum: 10c0/aefb57e8d1264486acc56e31eef3e0b3479a3dccd5f868136061d7baefc02c618071401d0fb71bd00536583702b1f8d7d95570c816069faec524b69dd0a37337
-  languageName: node
-  linkType: hard
-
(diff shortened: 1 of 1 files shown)


twrichards:
> Will this require the two-stage PR approach to removing the Wazuh security group? See the release notes from `@guardian/cdk`:
> 
> [guardian/cdk@`v61.5.0` (release)](https://github.com/guardian/cdk/releases/tag/v61.5.0)
> 
> Having said that, when I tried releasing [guardian/editorial-collaboration#24](https://github.com/guardian/editorial-collaboration/pull/24) to CODE without first doing this tweak, the riffraff deploy failed. So if you have already tested this on code, maybe there is some reason why it isn't necessary here?

good point/question!

just a guess, but I imagine that in your case the security group couldn't be deleted because there were active EC2 instances using it, but for Pinboard we only use EC2 very occasionally to spin up a box to use as a jump host to the [private] DB (as pinboard is otherwise serverless), the ASG is normally zero. so I think the PROD CD should be fine here 🤞

PR #395, opened 2026-03-26
Bumps aws-sdk dependencies to avoid vulnerable fast-xml-parser version

SHession:
## What does this change?

Ronseal change...

Worth a quick test on CODE.

Transitive vulnerable dependency is removed:
https://github.com/guardian/pinboard/security/dependabot/220


waisingyiu (approved):
👍

PR #404, opened 2026-07-29
Using @guardian/permissions-client

liamness:
## What does this change?

This removes logic relating to fetching permissions data from S3 and processing the result, instead using `@guardian/permissions-client` (which is defined in https://github.com/guardian/permissions).

I've deployed to CODE and checked that Pinboard functionality in Composer looks correct. Have also checked that the user refresher lambda is still running without issue (it is, but is simply reporting that no changes are required).

For anyone who is reviewing this, I'd be up for pairing on testing (given that Pinboard has collaborative functionality).

Validating that we have a shared implementation of permissions logic for use in Node.js backends


laurynas-karvelis-guardian (approved):
LGTM 🚀

dblatcher (approved):
looks good - tried it on CODE. Revoking my pinboard permission removed the button on Composer and after restoring the permission, the button came back.
Was abale to post messages to/from Composer and workflow.

